PR #187, opened 2025-11-24
refactor(resources): component consolidation phases 1-9

mfbayraktar:
## Summary
- move all single-use components to their feature folders as outlined in COMPONENT_CONSOLIDATION_PLAN
- add new app and feature-level component nests plus resources/internal restructuring
- wire updated component paths throughout pages and update barrels/imports

## Testing
- npm run build
- npx tsc --noEmit *(fails: pre-existing type errors in MachineRepositoryTable, SplitRepositoryView, CloneMachineManager columns)*


mfbayraktar:
Thanks for the thorough review and suggestions! I'll keep the createTruncatedColumn enhancements on the radar for a follow-up once the consolidation lands.

PR #508, opened 2026-06-10
fix(ci): resumable R2 transfers in promotion validation

mfbayraktar:
Unblocks main: Validate Promotion failed 3× consecutively on `e8c36c2d4` with mid-stream `IncompleteRead` from R2's S3 API (public CDN healthy, no Cloudflare incident — the CLI's default 10-way parallel range-GETs are the known trigger), and `cp --recursive` restarted the entire multi-hundred-MB edge tree on every retry, so the attempts never converged.

- Transfer profile: `max_concurrent_requests 3`, `multipart_threshold 64MB` (≈8MB debs become single GETs), `--cli-read-timeout 0`
- Recursive copies → `aws s3 sync` in the same retry wrapper: every outer attempt **resumes** (completed files skipped) instead of starting over

Once merged, the resulting main push re-runs the full pipeline with the fix, releasing the #504 content (charts + --detach) that is currently gated behind the red `CI Complete`. Merging via the queue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1159,15 +1159,26 @@ jobs:
           [[ -n "$AWS_ACCESS_KEY_ID" ]] || { echo "::error::R2_ACCESS_KEY_ID not set"; exit 1; }
           [[ -n "$R2_ENDPOINT" ]] || { echo "::error::R2_ENDPOINT not set"; exit 1; }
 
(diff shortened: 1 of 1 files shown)


mfbayraktar:
Superseded: the fix lands with #507 instead (commit dcd7da844 on 0610-3) per review preference.

PR #531, opened 2026-07-22
ci(ops): run the tutorial sequence inside ops-vm-provision (#424)

mfbayraktar:
Implements #424 (closed at implementation start per operator instruction — this PR delivers it).

## What

Adds the repo-tutorial sequence — `.ci/tutorials/tutorial-deploy-app.sh` (02) then `tutorial-forking.sh` (04), website order — INSIDE the existing `ops-vm-provision` job, recorded via `.ci/tutorials/record.sh` against the cluster the job already provisions. Cross-tutorial machine state is the point: renet#60 escaped because every command passed alone and only the 02→04 sequence broke (`repo fork --up` → EADDRINUSE). Per-command tests and the five static `check:ci-tutorial-*` gates are structurally blind to that seam.

- `timeout-minutes` 30→45. #423 removed recording over this timeout; raising it is the fix — a separate job would reprovision the cluster (~14 min) and double the VM bill.
- Worker IP exported from the existing `ops status` step (`ops status` lists every **configured** VM including absent ones, so the worker is selected by `id == 11`, never by array position).
- renet symlinked into `/usr/local/bin`: the tutorial preambles wipe `~/.config/rediacc/rediacc.json` (which carried `renetPath`), after which the CLI falls back to PATH resolution; the symlink also covers sudo `secure_path` for the `always()` `ops down`.
- Docker Hub auth written to `/root/.docker/config.json` on the worker over SSH (renet runs via `sudo`, so VM-side pulls read root's docker config; written as a raw file because docker isn't installed on the VM until `machine setup` runs inside the first tutorial's preamble). Uses the workflow's existing `DOCKERHUB_*` secrets.
- Casts uploaded as `tutorial-recordings-<sha>` artifacts, validation only — committed casts under `packages/www` remain the production artifacts. Recording scripts are untouched (operator decision: verbatim).

## Validated locally before CI (real --basic cluster)

- bash passes: 02 in 107s, 04 in 126s — one shared cluster, no reprovision between.
- record.sh passes (asciinema 2.4.0, same as CI's pip install): 02 in 102s, 04 in 124s, markers + idle compression OK.
- Docker-auth negative test: invalid `/root/.docker/config.json` → fresh pull through a repo daemon fails `unauthorized`; removing the file recovers the anonymous pull. Proves root's config is the read path and passwordless sudo works on the worker.

## Round log

- Round 1: `Quality / Deps` red — typescript-eslint 8.64.0→8.65.0 past the freshness window. Bumped the trio together at f4ad2b7d1 (workspace lockfile wedged incremental ERESOLVE; re-resolved via npm@10 uninstall+install). check:deps, check:lint, biome green locally.

## User-facing surface

None. CI-only change plus a renet-internal fix; no CLI commands added/removed/renamed.

## Submodule Branches

- private/renet: rediacc/renet#94 (`0722-1`) — flock-protect in-flight backup snapshots from the storage-maintain stale-snapshot GC (other session's completed work traveling with this wave's snapshot; `check:ci-renet` green locally).

---

## Wave 2 (7916a03c6): PR-flow redesign — draft PRs, pointer-bump fast path, Claude review

This branch now also carries the PR-flow redesign (the #424 tutorial-sequence payload above is staged work from a parallel session and lands with its next push; commit 7916a03c6 contains only the PR-flow scope):

- **Draft-until-green**: console PRs are created as drafts; `gh pr ready` is hook-gated on `CI Complete` = SUCCESS; `gh pr merge --admin` is banned outright (three new pre-bash guards + 41 harness cases). This PR itself rides as a draft.
- **Pointer-bump fast path**: `.ci/scripts/ci/detect-pointer-bump.sh` proves a push is content-identical to a full-CI-green baseline (pointer-only commits; tree-identical gitlink moves onto submodule main) and sets `pointer_bump_only`, which skips build-renet (cascading), migration-test, stripe-sandbox, package-tests, and ops-tests; `assert-ci-complete.sh` accepts those skips only under the flag (new gate test `test-ci-complete-tiers.sh`, 8 assertions). Detector proven against PR #530's real pointer commit plus synthetic negatives.
- **Gemini removal**: trigger/wait scripts, `poll_with_watchdog`, `MAX_GEMINI_REVIEWS`, the shellcheck shared-constants guard, and the `no-gemini-review` label entry are gone (the bot is sunset upstream). The `review-gate` job keeps the bot-agnostic resolved-threads + review-comments checks.
- **Claude review**: `claude-review.yml` reviews exactly when CI is green AND the PR is non-draft (`workflow_run` success + `ready_for_review` entries, marker-comment dedup, incremental follow-up reviews that never re-block already-reviewed code); `claude-mention.yml` answers @claude from write-access users. Note: `workflow_run` activates only once these files reach the default branch, so the first automated review fires on the next wave after this merges.
- **worklist.py v3**: a `[>]` lease state so stop-hook enforcement coexists with delegated background work.
- **Docs**: pr-babysitter, pr-babysit, pr-merge, ci-gates.md, CONTRIBUTING.md rewritten for the new flow (CONTRIBUTING previously documented required checks that do not exist).

Local verification for wave 2: hook harness 41/41, full quality-gate suite green, workflow compliance battery + `check:actions` green, detector and review-gate script dry-runs against live PR history.

### Wave 2, round 2 (c18dadc71)
Round-1 red was `Quality / Renet`: klauspost/compress v1.19.0 -> v1.19.1 upstream release tripped the go-deps freshness gate (time-drift, not wave-2 code; compress became a renet direct dep in the embed wave). Fixed with operator permission: renet ead3e8c bumps compress (rediacc/renet#94), console pointer follows. Also rides: worklist.py owner-tag parsing fix + dead-session cleanup (in-place tombstones under flock).

### Wave 2, round 3 (251d59136): consolidation + Renet (Full) un-walled
Operator-sanctioned consolidation of all parallel workstreams: executor daemon (command + services/executor/daemon/, registry/contract/i18n already wired), the #424 tutorial-sequence payload, the clean-stdout cluster (console rdc.sh + renet 7cf0ea9), and the darwin cross-compile fix (renet 305b452: three cross-platform pkg/prune symbols moved out of the linux-tagged file -- the round-2 Renet (Full) red; all six GOOS/GOARCH targets proven). CI optimization: embed-asset staging cache keyed on embed-assets.lock.json + native reconstruction via zstd on cache hits (proven byte-identical locally) + a ~2-min cross-compile smoke job. Also: orphan executor-daemon i18n key removed (command is deliberately hardcoded English). Local batteries green: full check:ci-renet, CLI unit suite, contract + all i18n gates, workflow compliance.

### Wave 2, round 4 (e35f91574)
Round-3 reds, all three fixed: smoke-job checkout scoped to the renet submodule only (least-privilege token 404'd on account/elite under submodules:true); command-help-coverage gets a HARDCODED_ENGLISH exemption for executor-daemon (the orphan-key gate and the coverage test demanded opposite catalog states for a hidden hardcoded-English command); run-sequence.sh mapfile replaced with the while-read minimal-CI pattern. Full CLI suite re-proven green AFTER the changes.

### Wave 2, round 5 (0e064a6c1)
Round-4 red: Quality/Format on three executor-daemon files from the snapshot (biome). Formatted from repo root, daemon tests re-run green.

### Wave 2, round 6 (c299adccf)
Round-5 red: Quality/Lint no-unused-vars on TRANSPORTED_CALLBACK_KEYS in daemon protocol.ts; fixed by using the constant in toWireOptions (author intent), tsc + eslint + 66 daemon tests green.

### Wave 2, round 7 (501c89dae)
Round-6 red: OPS Provision died at the payload's new rustfs step -- the bridge VM has no docker under --skip-orchestration (local record flow orchestrates, CI deliberately does not; classic first-live-run gap). Fixed with `renet ops docker install 1` before rustfs start; workers still get docker via machine setup in tutorial preambles, preserving what the sequence proves.

### Wave 2, round 8 (3f00e75af)
Round-7 red: same OPS step, one layer deeper -- ops docker install expects renet deployed on the VM (orchestration's job). Fixed: ops service install-renet 1 before docker install (SCP + versioned install, the exact path the error named).

### Wave 2, rounds 9-10 (e4ea6ef88)
Round 9 attempt 1: Migration + OPS green (all prior fixes hold); 4 E2E reds = infra flakes (workers: apt index outage; ceph trio: node-22 setup OOM-killed -> ceph absent -> blank-health timeout). Attempt 2 cleared 3 of 4 and exposed the one REAL regression: 13-ceph-datastore asserted pretty-printed JSON ("mounted": true) against renet's new compact one-line output. Fixed by parsing (the documented consumer contract); swept the class, single site.

### Wave 2, round 11 (1f88cadb2)
First live Claude review (fired on the ready-flip; the ready_for_review entry runs from the PR branch, so no bootstrap wait needed) hit error_max_turns at 30 with zero comments posted on this 90+-file diff. Raised to 50 turns + breadth-first prompt strategy (rank by risk, post inline as confirmed, always reserve budget for the summary). Review re-fires via a draft/ready toggle once this round is green.

### Wave 2, round 12 (aab177a6a)
Review turn budget now scales with diff size (50 / 80 at 5K+ / 140 at 30K+ lines, gate-computed from additions+deletions; thresholds proven on live PR data) and the initial prompt carries mega-diff coverage rules: area-by-area scoped reads, generated bulk deprioritized, explicit coverage map in the summary, /code-review ultra recommended when coverage is materially partial. Supersedes round 11 (flat 50).

### Wave 2, round 13 (cf9f9f1f6)
Review runs 1-2 root-caused: the Bash(...) allowlist patterns (spaces inside) did not survive claude_args splitting -- the reviewer had zero gh access (run 2: "success", 12 turns, 36 permission denials, nothing posted, and the marker then wrongly stamped the SHA). Allowlist moved to the action settings JSON; --mark now refuses to stamp unless output actually posted (guard proven by live negative control); the false marker was deleted so the SHA is retryable.

### Wave 2, round 14 (3774d4810)
The intermittent tutorial ssh failures root-caused BY LOCAL REPRO: ssh2's ed25519 encoder mis-pads ~0.6% of generated keys (private blob 3 bytes short); OpenSSH rejects them at connect time as "error in libcrypto". Only repo-ref connects use generated keys, which is why machine-ref connects never failed. Fix: validate-and-retry using ssh2's own parseKey (rejects exactly the malformed outputs). Proven: 3/500 bad unfixed -> 0/500 fixed against ssh-keygen -y; 78 consumer tests green.

### Wave 2, round 15 (3391ecfaf)
Round-14 Quality/Deps red: the fix was the operator's own uncommitted deps work (otel 0.221 trio, hono 2.0.11, biome 2.5.5 + syncpack pin updates + lockfile). Committed with authorization as the deps cluster ONLY: root package.json split at hunk level via the index so the in-flight rdc-sh-env gate wiring stays uncommitted with its workstream; lockfile npm@10-validated.

### Wave 2, round 16
Round-15 red was the SAME deps class in private/account (sweep miss owned): operator-authored bumps committed as the deps pair only -> rediacc/account#66 (branch 0722-1), console pointer follows. Two of my own errors caught and repaired en route: a false "COLD REIFY OK" (marker chained off tail, not npm -- the real run then exposed a GENUINE npm-11 lockfile drift: esbuild@0.28.1 entries missing, reconciled with npm@10 and proven by cold npm ci exit 0) and a detached-HEAD account commit (branch created at the commit, nothing lost).

## Related PRs
- https://github.com/rediacc/renet/pull/94
- https://github.com/rediacc/account/pull/66

### Wave 2, round 17 (9d9c137e0)
Round-16: Security = job-timeout flake (greened on retry). The retry exposed the account bump's one real casualty: better-sqlite3 v13.0.1 ships no node-22 prebuilt, and the sandbox image has no gyp toolchain -> Stripe Sandbox red. Reverted to ^12 in account (c34d76a, npm@10 cold-proven) + console blocklist BLOCKER entry until upstream prebuilds publish.

### Wave 2, round 18 (0c9bffd94)
Review attempt 3: mark-guard fired correctly (job honestly red, no lying marker). Forensics: SDK-path denials persist (29) despite settings-JSON permissions that byte-identical local runs honor; 43 turns/$4 of reading with every posting tool denied. Fix: track_progress -- the ACTION posts the report comment via github_token; model-side posting becomes best-effort.

### Wave 2, round 19 (3d4b81dab)
Round-18 red: Quality/Security killed at 18min by ubuntu-slim's hard ~15-min runner cap (runtime grew 8->12->18 across the wave; 2 kills in 3 runs; the new watchdog proven uninvolved -- its log says the run continues unwatched pre-merge). Moved to ubuntu-latest; the runtime growth is a logged follow-up.

### Wave 2, round 20 (9a456f1d5): THE REVIEW WORKS -- and its findings land
Attempt 4 SUCCEEDED: 7.3KB report posted by the action (track_progress), marker honestly stamped by the guard. Its High finding was a real bug in the mark-guard itself (unpaginated comment reads -> permanent re-review loop on >100-comment PRs) -- fixed with pagination, proven against this PR's own live comments; the detect-pointer-bump muted-error minor also fixed. The three executor-daemon findings (config resurrection race, double-spawn token desync, hung-daemon no-timeout) are handed to the daemon workstream.

### Wave 2, round 21 (374b55795)
Operator request: the per-review marker comment now carries cost transparency -- total USD, model, turns, duration, and in/out/cache-read/cache-write tokens, parsed from the action's execution file (object and stream-array forms proven on real fixtures; missing file never blocks marking). Rides with the round-20 review-findings fixes; the follow-up review after this green exercises delta scoping AND the cost display together.

### Wave 2, round 22 (f82fad47a): submodule reviews via reusable workflow
Operator-requested: claude-review-reusable.yml is the single source of review mechanics; console is a thin caller (triggers + CI Complete green gate unchanged); renet (5e88610) and account (9bcb1c2) get thin callers -- review on PR open + workflow_dispatch, NO green gate (they have no PR CI; renet's workflow registrations are stale relics, validation lives in console CI). Gate gains REQUIRED_CHECK + dispatch head resolution, all three modes proven live. Prompt: hand-maintained configs get a sanity pass; pointer moves declared reviewed-in-their-own-PRs. OPERATOR ACTION: set CLAUDE_CODE_OAUTH_TOKEN on rediacc/renet + rediacc/account (elite deferred: no active branch).

### Wave 2, round 23 (6bc9c4f7f)
Round-22 attempts died of ENOSPC twice on different jobs (ops: runner worker crashed mid-sequence, no logs; stage-artifacts: archlinux pull for pacman metadata). Structural fix: the ct-tests.yml preinstalled-bloat removal (already at 5 sites there) added to both tight jobs.

### Wave 2, round 24 (146881e38)
Operator request: Gemini-style line-anchored review comments with severity badges. The reviewer emits a machine-readable findings block (cap 20, severity-ranked) inside its report; a deterministic step posts them as **[SEVERITY]** inline comments via github_token (no model-side tools -- the report text is the only channel proven to escape the action sandbox). The comments create real threads the resolved-threads gate enforces. Proven: extraction+ordering on fixture, live post+delete on this PR's own diff.

### Wave 2, round 26 (fce9dbf40): overnight consolidation (operator /goal)
Operator's config-universe + licensing wave committed in full per the standing goal: console 230 files (config remote/handoff, subscription fingerprint, rdc-sh-env gate now complete with its script, docs), renet fac725e (licensing delegation/preferred/fingerprint, 29 files, FULL check:ci-renet green), account ba9e06c (54 files, ConfigRemote portal + handoff). Local pre-push battery all green: rdc-sh-env gate, tsc, cli-contract, command-tree, CLI suite 2133/2133, check:i18n.

## Round 27 (2026-07-23)

Round 26 failed 25/25 real jobs on ONE shared root: `private/account`'s committed
package-lock.json was out of sync with its package.json (config-universe snapshot
brought deps whose transitive esbuild moved to 0.28.1; no lockfile companion), so
`cd private/account && npm ci` in install-deps.sh hit EUSAGE in every job.
Reproduced locally (npm@10 dry-run in the submodule), fixed by npm@10
`install --package-lock-only`, verified by a real cold-cache `npm ci` reify in a
clean-room checkout (exit 0). Swept the class: account `e2e/` and `web/` lockfiles
verified in sync. Account 2be0d53, console pointer bump only.

## Round 28 (2026-07-23)

Round 27 (first real CI exercise of the operator wave after the lockfile fix): 3 reds,
all reproduced locally before fixing. (1) Quality/i18n — stale search indexes,
regenerated all 14. (2) Caption Sync — one transient fetch failure out of 180 while
the asset was live with matching sha256; checker now retries network/5xx 3x
(full 180-combo live rerun green). (3) Unit — new server-binding test blew the 5s
vitest default on a starved runner; testTimeout 30s, suite 2133/2133 green.

Also lands the review-report reply gate (operator request): the review's summary
report (issue comment) now requires a substantive reply referencing it by id —
`check-review-report-replies.sh`, third step of review-gate. Proven live on this
PR's own report (fires unreplied → passes after the posted per-finding reply).

## Round 29 (2026-07-23)

Round 28: 3 reds -> 1. Search-index regen, caption-fetch retry and the vitest
timeout all held; the only failure was chain-parity catching the NEW report-reply
gate as a workflow-invoked script missing from `npm run ci`. Same class as its two
siblings (needs a live PR + GitHub API), so it joins them in `.ci-chain-exempt`
with a BLOCKER reason. Reproduced locally first; chain-parity green (8 exempt),
suppression liveness green (80 entries, 0 findings).

Also in this push: the follow-up reviewer now consumes report replies (verifies
"fixed in <commit>" claims against the delta, does not re-raise tracked
deferrals), a bootstrap fallback so reviews work before the review system reaches
main, and the Claude hook harness is finally CI-gated (new gate test picked up by
run-all.sh; shellcheck/shfmt scopes extended to .claude/hooks). The merge hook now
also enforces report replies at merge time -- the one window ci.yml structurally
cannot cover, since required checks are per-commit and a report posted after green
can never re-red them.

## Round 30 (2026-07-24)

Round 29: 1 red — Stripe Sandbox, the first round in this wave where quality went
green far enough for that job to actually run. It exposed a real defect from the
operator's SQLite migration: better-sqlite3 13.0.1 ships no node-22 prebuilt, so
the account image build fell back to node-gyp and died on missing Python.
Verified locally this is not a musl quirk (node:22-alpine AND node:22-slim both
fall back). Both installing stages now get python3/make/g++; the runtime stage
installs them virtual and deletes them in the same layer, so only the compiled
better_sqlite3.node ships. Proven end to end locally: compose build succeeds, the
module loads and runs SQL inside the shipped image with the toolchain absent,
ci-start-account.sh exits 0, container healthy, /health 200.

Also dropped the better-sqlite3 entry from .deps-upgrade-blocklist: its reason
("image has no python/gyp toolchain") is now false and latest == installed ==
13.0.1, so it blocked nothing. Leaving a stale reason is exactly what
suppression-liveness exists to catch. Both deps gates green after removal.

## Round 31 (2026-07-24)

Round 30: everything green except one job — `OPS Provision (macos-intel)` hung on
`Test: rdc ops down` for ~50 minutes, taking the run to "cancelled" and CI Complete
red. The same job passed in 4 minutes in round 29, so this was the hang itself, not
a regression from this branch.

It turned out to be a real product defect rather than CI flake. `runOpsCommand`'s
15-minute timeout DID fire, but it only sent SIGTERM to renet. renet spawns the VM
process (QEMU/KVM) as its own child inheriting our stdout/stderr pipes, so the
grandchild kept those pipes open: `close` never fired and the live handles kept the
event loop alive even after the promise rejected, so the CLI hung well past its own
timeout. The timeout path now escalates to SIGKILL after a 10s grace, stops reading
the inherited pipes, and unrefs the child.

Regression test reproduces the real topology (parent ignores SIGTERM, spawns a
grandchild holding the pipes) and is proven live BOTH ways: the pre-fix shape
reports the timeout then hangs until externally killed, the fixed shape exits 0.
CLI suite 2134/2134, tsc clean.

Still open (reported, not fixed): why macOS QEMU teardown wedges at all. It is now
bounded and surfaced as an error instead of consuming the run.

## Round 33 (2026-07-24)

Round 32: two reds, both real.

**Deps** — the freshness gate demanded two bumps that published after the previous
round's local run (time drift, not a code problem): claude-agent-sdk 0.3.218 and
knip 6.29.0, taken via the gate's own `--upgrade` path with the lockfile verified
in sync under npm@10.

knip 6.29 then surfaced 4 configuration hints (treated as errors). Fixing them
properly retired two stale suppressions instead of papering over them:
- `ignoreBinaries`: 6.29 now resolves `which` and `security` natively, so both
  entries were dead. Dropped, and the BLOCKER comment retuned — it already records
  this same churn for systemctl/cmd.exe/powershell.
- `project` globs were missing compiled extensions, so knip was not following
  imports into www's `.mdx` or account-web's `.css`. Added both.
- Consequence worth stating: making `.css` visible let knip resolve
  `@import 'tailwindcss'` in app.css, which retired the `tailwindcss`
  ignoreDependencies entry whose stated reason was literally "CSS references are
  invisible to knip". That reason is now false, so the entry is gone.

(Commit 1d0464f04's message is missing those backticked terms — bash ate them as
command substitution in a `-m` string. The diff is correct; this section is the
accurate record. Not force-pushing over published history to fix prose.)

**PR Description** — the gate requires the body to be edited within 30 minutes of
the newest commit, and I pushed round 32 without refreshing it. Process fix: the
body refresh now happens immediately after each push, not before it.

Gates green locally before push: lint:unused (0 hints), check:ci-knip-blockers,
check:deps, check:lint, check:format, check:ci-lockfile, suppression-liveness.

## Round 34 (2026-07-24) — CI reached GREEN

Round 33 attempt 1 failed only on a transient Docker Hub auth timeout pulling
node:20.20.2-alpine inside the repo's own dockerd (postgres pulled fine). Manual
rerun (the watchdog cannot auto-retry pre-merge, see below): **attempt 2 = 97
success / 19 skipped / 0 failed, CI Complete SUCCESS.**

With CI green the ready-toggle fired the review — and it died instantly as
`startup_failure` with no job and no log (run 30064158589). Root cause: **a called
workflow cannot exceed the CALLER's permission grant.** The callers default to
`contents: read`, while the reusable's job needs `pull-requests: write` (inline
comments), `issues: write` (marker upsert) and `checks: read` (required_check
lookup). For a normal job, workflow-level permissions are only a default the job
may widen; for a reusable CALL they are a ceiling. So the reusable conversion had
never successfully started anywhere — the last green review predates it and ran as
a single-file workflow. Fixed on all three callers (console, renet, account),
scoped to the calling job.

Two pre-merge facts worth recording, both structural rather than bugs:
- `workflow_run`-triggered workflows only ever run the DEFAULT BRANCH copy, and
  neither `claude-review.yml` nor `watchdog-monitor.yml` exists on main yet. So the
  review can only be fired via the `ready_for_review` path until this merges, and
  the CI watchdog is inert — `dispatch-watchdog.sh` says so honestly
  ("continues UNWATCHED"), which is why every transient this wave was hand-diagnosed.
- The review's bootstrap fallback (added earlier this wave) is what will let it run
  at all, since `.ci/scripts/review/` is not on main either.

## Round 35 (2026-07-24) — E2E zero-skip contract

Operator flagged 48 skipped tests per E2E Workers leg (x5 OS). Root-caused locally
with `playwright --list` (no VMs): the base playwright.config.ts collects
NN-*.test.ts recursively but its testIgnore listed only ceph, so it pulled in the
kube (33) + migrate (5) suites and skipped them, plus 10 destructive ops-lifecycle
tests self-skipping via test.skip(!!CI). Those suites actually run in the dedicated
K8s/Ceph/Multinode/Migrate jobs (all verified 0-skip in the last green run).

Fix (four operator-ruled parts): base testIgnore now excludes every topology subdir
(kube/migrate/ceph/ops-lifecycle); destructive VM lifecycle/reset moved to
tests/ops-lifecycle/ (covered by the OPS Tests job), Parallel Execution kept;
run-e2e.sh --fail-on-skip fails any job with skipped>0 (fails closed without the
reporter sentinel), wired into all 7 E2E jobs; scripts/check-e2e-skip-hygiene.ts
statically forbids the regression (ci-quality E2E Coverage job + npm run ci).

Proven locally: --list kube/migrate 38->0; instrument test fires-on-skip /
passes-on-zero / fails-closed; full e2e-tests lint+tsc green. "Verify then enable
all" satisfied by the green run's dual-signal per-job skip counts (all specialized
jobs already 0-skip). This round is the first with the gate live across all 7 jobs.

---

## Wave 4 (3e24238ca, da10dc833): CI quality-phase reorganization

Measured on run 30064285223: the quality phase burned **4307s of runner time,
3042s (73%) of it setup rather than gate work**, against a hard 20-slot free-tier
cap that sat at exactly 20/20 for the whole phase.

- **`filter: blob:none` on all 9 `fetch-depth: 0` sites.** `git cat-file` says
  5735 MB of this repo's 5741 MB pack is blobs (pre-migration video history);
  commits plus trees are 5.7 MB. Proven live: `Quality / Branch` 315s -> **22s**,
  `Submodule Branches` 301s -> **32s**, `i18n` 478s -> **198s**.
- **New `./.github/actions/setup-workspace`** caches `node_modules` (200 MB,
  untars in 2.5s) instead of `setup-node cache:'npm'` + `npm cache clean --force`
  + `npm ci`, which was deleting its own 396 MB cache and re-downloading 1096
  packages in every job. All six `install-deps.sh` callers moved over, E2E legs
  included.
- **33 quality jobs -> 10 lanes**, grouped by setup class. Every gate step
  carries `if: !cancelled() && steps.setup.outcome == 'success'`, so a lane
  reports ALL its failures in one round and a failed install cannot cascade.
  Verified: zero gates lost; 86 gate steps across 8 multi-gate lanes guarded.
- **Watchdog drains before cancelling.** `WATCHDOG_NO_RETRY_PATTERNS` is
  `Quality,Review Gate` matched with `String.includes`, so a quality failure
  force-cancelled the run instantly and killed the other lanes mid-flight -- that,
  not job count, is why gates surfaced one failure per round. Still no retry and
  no AI, but the cancel now waits for sibling no-retry jobs and fires once with
  the full roster. Review Gate keeps its instant kill.
- **`ubuntu-slim`'s 15-minute cap is now enforced.** New CHECK 3 in
  `check-workflow-gates.sh` requires every slim job to declare
  `timeout-minutes <= 14`, with a both-ways fixture test. It caught two live
  lies: `housekeeping`'s `cleanup` (60) and `stripe-cleanup` (30) declared
  timeouts the runner can never honour, so a long run was silently CANCELLED at
  15 rather than timing out. Both moved to `ubuntu-latest`.
- **`Migration Test` moved into `Tests + Infra`.** Not a rename: `full_suite`
  moved from the caller's `if:` into a `ct-tests` input and now gates each test
  job individually, so migration-test still runs on push-to-main (the path that
  deploys those D1 migrations). Cost: it starts after `build-docker-fast`
  (~10 min in) instead of ~2 min in.
- Deferred git-history rewrite filed as #532 with the object-size measurements;
  the CLAUDE.md paragraph is now a pointer.

First-run result (cold node_modules cache): quality runner time **1847s vs
4307s**, wall **412s vs 478s**, 10 slots instead of 20. `da10dc833` fixes the one
red: `check-commands.sh` scans `.sh` as bash with no heredoc scoping, so CHECK
3's Python `timeout = ...` read as a bash `timeout` invocation (which, with
spaces, it genuinely would be). Renamed the local rather than teaching the
scanner to skip heredoc bodies, which would blind it to
`ssh host <<'EOF' ... timeout 5` cases.

Not verifiable on a PR: the watchdog drain. `watchdog-monitor.yml` is not on the
default branch yet (this branch carries it), so `workflow_dispatch` 404s and the
chain fails open -- the drain first runs for real after this merges.

## Round 36 (2026-07-24) — review post-steps fix

First successful end-to-end review (permission fix + bootstrap proven; report +
findings block + cost line all posted) had one residual bug: "Post inline findings"
hit ENOENT on the gate script and the marker step skipped, so 0 inline comments +
no dedup marker. Cause: the anthropics action resets the workspace's untracked
files mid-run, deleting .review-scripts after the gate used it (not a checkout wipe
— clean:false held). Fix: stage the trusted .ci tree into $RUNNER_TEMP right after
the gate and run --post-findings/--mark from there, outside $GITHUB_WORKSPACE.
The gate (security decision) still runs from .review-scripts pre-wipe; the staged
copy is whatever fed the gate, so a PR tree can't forge the marker. Verified
locally (actionlint, workflow gate, relocated-script loads + exits 0). Fixed
autonomously per the standing overnight /goal; operator's clean:false/blob:none
preserved.

## Round 37 (2026-07-24) — all 4 review findings fixed

The first successful end-to-end review found 4 real defects (2 High, 2 Medium),
including flaws in this wave's own fixes. All fixed in 21925450d, verified locally
(cli+shared tsc, eslint, shell gates, hook harness 48/48, CLI suite 2134/2134):
- F1 (High) hook quote-strip bypass -> shared command-scan.sh (unwraps sh -c/eval,
  strips heredocs, matches flag=value) + 7 adversarial harness cases.
- F2 (High) shell injection via server-controlled storageKeyId -> execFileSync arg
  arrays across all backends + ValidatingStorage allowlist choke-point + schema regex.
- F3 (Med) ops SIGKILL never fired (process.exit preempted the unref'd timer) ->
  detached process-group kill; test now proves the real grandchild dies.
- F4 (Med) review-gate lost fallback guards + fence truncation -> restored graceful
  degradation on all paginated reads + awk extraction; BONUS review-found
  last_marker_sha tail-before-sed bug (dead dedup) fixed.
Report reply posted (report-reply gate passes). Review Gate should clear this round.

## Round 39 (2026-07-24) — biome format fix + actions freshness

Round 37's real defects were all fixed (4 review findings); the only remaining red
was a `check:format` failure (biome wants single quotes on two plain PowerShell
script strings in the F2 secure-storage.ts fix) — fixed directly by the operator
(1a6f8f9d7), verified format/tsc/eslint clean, no CI round needed.

Then `npm run check:actions` flagged docker/login-action v4.5.0 -> v4.5.1 (34 pins).
Resolved the new SHA and rewrote every pin with the gate's own printed command
(17ad9ff0c); check:actions green (14/14), actionlint clean, workflow gates green.

PR description freshness gate: I missed refreshing the body across those two
pushes — fixed now, this edit covers both.

## Round 40 (2026-07-24) — F1 follow-up + PR-description gate (again)

Round 39's Claude review confirmed F2/F3/F4 all correctly fixed, and the marker
upsert + $RUNNER_TEMP staging worked end to end for the first time (dedup marker
updated in place with fresh cost, not duplicated). It found ONE real High finding:
my F1 fix still had a bypass via flagged shell wrappers (`bash -lc`, `sh -eu -c`,
`bash -eux -c`) — the shell name and `-c` had to be immediately adjacent, and the
prose-strip step separately erased the same quoted payload, so the guardrail hooks
exited 0 while the shell genuinely executed. Verified live before fixing (the hook
returned exit 0 on the bypass), fixed by tolerating intervening flags on both the
extraction and prefix-strip regexes (16865258f). Harness 51/51, reproduced the
bypass blocking afterward. Report reply posted, report-reply gate passes.

Round 40's only red was the PR-description staleness gate again — I pushed
16865258f without refreshing the body. Fixed here. Standing correction: I keep
dropping this step after security/hook fixes specifically; treating the body
refresh as part of every push from now on, not a separate afterthought.

## Round 41 (2026-07-24) — token-scanning replaces the flag-shape regex

Round 40's review verified LIVE (actual grep runs) a third gap in the F1 hook fix:
GNU long options (`bash --posix -c`, `bash --norc -c`) and value-taking short
options (`bash -o pipefail -c`) still bypassed the guardrail hooks entirely. Rather
than add a fourth flag-shape case, replaced the regex with token-scanning
(bd7eabcd4): an awk pass that skips any intervening token and asks only "is this
the -c selector" — closed and enumerable, unlike flag syntax. Also removed the
structural flaw behind all three rounds: an extraction regex and a separate
prefix-strip regex had to independently agree on shape; the new function emits the
payload directly. Verified via benign placeholder commands (not the literal banned
string, which the fixed hook now correctly intercepts). Harness 54/54 (was 51, 48).
Report reply posted, gate passes.

## Round 42 (2026-07-24) — GraphQL retry/validation fix

Round 41's Review Gate crashed with a raw jq parse error (`Invalid numeric literal`)
and a broken-pipe message — not the report-reply gate. Root cause read from the
code: `gh api graphql` can exit 0 with a truncated/malformed body, the existing
command-exit fallback never catches that, and the immediate `.errors` check
silently swallows the parse failure too (looks identical to "no errors field"),
falling through to a confusing crash instead of a clear exit. Fixed (24b034367):
retry the fetch up to 3x validating JSON parseability each time, sentinel fallback
only after exhausting retries. Verified both directions — live query still passes,
a fake gh forcing malformed JSON drives the full retry-then-clean-exit path.

The other round-41 red, Build (CLI) / CLI (win-arm64)'s 403 on artifact listing,
is a transient Actions infra error unrelated to this push; a fresh run resolves it.

## Round 43 (2026-07-24) — basename fix for path-qualified shells

Round 42's review verified live (grep -x probe) a fourth gap in the same detector:
the shell-name test required an exact bare token, so `/bin/bash -c`, `./bash -c`
bypassed it the same way earlier flag shapes did. Fixed per the reviewer's own
direction (bc2b0890e): match the token's basename instead of enumerating path
prefixes. Verified with benign placeholders across all path forms plus the
plain-script-invocation negative control. Harness 56/56 (54->56). GraphQL retry
fix confirmed correct by the reviewer, no action needed. Report reply posted,
gates verified passing immediately.

## Round 45 (2026-07-24) — quote-strip fix; review count now capped at 3/PR

Round 44's review traced by hand a fifth bypass in the same function: the basename
strip only removed a path prefix, not surrounding quote characters, so
`"/bin/bash" -c '...'` still bypassed the wrapper detector (last `/` lands inside
the quotes). Fixed (d6aec6fa0) by also stripping quotes on the name-only copy.
Verified with benign placeholders across all quoted/unquoted/path forms plus the
negative control. Harness 58/58 (56->58).

Operator directive: capping automated review passes at 3 per PR going forward
(implementation follows this push).

## Round 46 (2026-07-24) — review cap implemented

Per operator directive, automated review passes are now capped at 3 per PR
(d34f9c9c6). New review_report_count() counts finished reports (same signature
the report-reply gate uses) and the gate refuses to fire once that hits 3 —
CI itself is unaffected, only the review pass stops. Verified live against this
PR (6 existing reports): dry-run correctly emits go=false with the cap reason.
Boundary logic verified both directions. This PR has already exceeded the new
cap from before it existed, so no further automated review will fire here going
forward — consistent with the operator's intent.

## Landing (2026-07-26) — submodule PRs merged, pointers bumped

Operator invoked the land path. Submodule-first, squash, branches kept:

- `rediacc/renet#94` merged -> `88503faf3` on renet `main`
- `rediacc/account#66` merged -> `16e0eeb78` on account `main`

Squash leaves those commits off the branch tips, so this push (`d0ab76e95`)
moves the console pointers onto the merge commits. Both squash trees verified
byte-identical to their branch tips (`git diff --stat <tip> <squash>` empty),
so this is a pointer-bump-only change: the `Submodule Branches` gate passes via
the ancestor-of-main path, and CI takes the pointer-bump fast path (quality +
review-gate + ci-complete only, builds legitimately skipped).

Note on the review marker: it sits at `bc2b0890e` while head is `d34f9c9c6`.
The two intervening commits are the quote-strip hook fix and the review cap
itself. The marker cannot advance by design — the operator's 3/PR cap landed in
`d34f9c9c6` and this PR already carries 6 reports, so the gate correctly refuses
to fire another pass. Stated here rather than worked around.

## Landing round 2 (2026-07-26) — two reds fixed without waiting out the run

Operator asked to fix the in-flight run's failures rather than wait for the
rest of it. Both reproduced locally first.

**Quality / Security** (`audit.sh`): postcss `GHSA-r28c-9q8g-f849`
(sourceMappingURL path traversal, arbitrary `.map` disclosure) is a PRODUCTION
advisory and therefore zero-tolerance -- root override `>=8.5.10` -> `>=8.5.18`,
tree resolves 8.5.23. brace-expansion `GHSA-mh99-v99m-4gvg`: the v5 edge is
fixed (`5.0.7` -> `5.0.8`); the six eslint/minimatch@3 edges at 1.1.16 have no
upgrade path -- the advisory patches at 5.0.8 only, the 1.x line ends at 1.1.16,
and forcing v5 there crashes eslint (tested: `require('brace-expansion')` on
5.0.8 returns `{EXPANSION_MAX, EXPANSION_MAX_LENGTH, expand}` and calling it
throws, which is exactly `minimatch@3.1.5:10/:271`). Dev-only, expands our own
lint globs -> allowlisted with that evidence.

**Quality / Content** (`check:deps`): time drift. claude-agent-sdk 0.3.220,
hono 4.12.32, and on the account side aws-sdk client-s3 3.1095.0 + wrangler
4.114.0 (landed as rediacc/account#67, pointer -> `eb1604076`).
`@playwright/test` added to `.deps-upgrade-blocklist`: `playwright` was already
held there for browser revalidation but its sibling was not, and
`@playwright/test@X` pins `playwright@X` exactly -- the gate was demanding the
same deferred bump through the back door.

**Also in this push** (`e2fc3ff88`), a 6th finding in the gh-pr guardrail hooks,
this one caught by it firing wrongly against my own merge: all three hooks
parsed `--repo`, the PR selector, `--auto`, `--draft` and `--undo` from the
WHOLE bash line. That both cross-attributed fields between sibling invocations
(the false positive I hit) and left every invocation after the first
unexamined -- including `gh pr ready --undo 1; gh pr ready 531` reading as an
always-allowed undo, a clean bypass of the green gate. Fixed at the scope with
`hook_gh_pr_segment`/`hook_target_repo`/`hook_pr_selector`; all three hooks now
judge each invocation on its own. Harness 58 -> 61, each new case proven to
fail against the pre-fix code.


mfbayraktar:
Re: review report 5059418479 — per-finding response.

**1. High — mark-guard misses recent comments (unpaginated `per_page=100`): FIXED** in `9a456f1d5` (`fix(review): paginate the mark-guard`). Both queries now use `--paginate` with per-page `--jq` id-streaming + `wc -l`, proven against this PR's own >100-comment history. The secondary note (an unrelated `@claude` mention satisfying the guard) is accepted as a known-narrow window: the guard is a liveness check against *silently empty* reviews, and scoping it to a run id would couple it to action internals we've already found unstable; risk is a false *mark*, which the next green-CI cycle's follow-up review corrects.

**2. Medium — `config-file-storage.ts` resurrection race (`loadUnlocked` ENOENT → empty config inside the lock): CONFIRMED REAL, deferred + tracked** in the repo worklist for the daemon/config workstream. Planned fix is exactly the reviewer's shape: make `loadUnlocked` honor `createIfMissing` so `updateState`'s no-create contract holds inside the lock, not only at the pre-check.

**3. Medium — daemon double-spawn token desync (`claimSocket` existsSync race): CONFIRMED REAL, deferred + tracked** same worklist. Fix direction: atomic claim via `O_EXCL` lockfile (or bind-first-then-write-token ordering so the token on disk always belongs to the process that owns the socket), plus token re-read on `tokensMatch` failure before falling back.

**4. Low-Med — client lacks a hung-daemon timeout: CONFIRMED, deferred + tracked** with 2-3. An inactivity timeout on the frame reader (reset per frame, generous enough for long SSH executes) with a stale-daemon diagnostic + direct-execution fallback.

**Minor notes:** `detect-pointer-bump.sh` dead `xargs -r` error message — already fixed in the same commit as finding 1 (`9a456f1d5`, two-step lookup so the intended `no_fast_path` message actually fires). `protocol.ts` frame-buffer growth — accepted for now (same-UID trusted peer), will get a size cap alongside the finding-4 timeout work since both touch `createFrameReader`.

Findings 2-4 are in the executor-daemon surface that ships in this PR but is exercised behind explicit opt-in; they are tracked as follow-up hardening rather than blockers. This reply satisfies the new `check-review-report-replies.sh` gate (report-level accountability), which this PR also adds.

mfbayraktar:
Re: review report 5068756345 — all four findings FIXED in commit 21925450d (verified locally: cli+shared tsc, eslint, shell lint/format, hook harness 48/48, full CLI suite 2134/2134).

**F1 (High) — hook quote-strip bypass: FIXED.** New `.claude/hooks/pre-bash/lib/command-scan.sh` shared by all three hooks: it unwraps `sh/bash/dash/zsh -c` and `eval` payloads and scans them at command position (closing the wrapper bypass), matches the flag in every form incl. `--admin=value` and assignment/indirection, and strips heredoc BODIES (which also fixed a real false-positive that fired on a worklist write). Added 7 adversarial + heredoc cases to test-hooks.sh (wrapper, eval, flag=value, variable-indirection, wrapped create) — harness 48/48. You were right that the harness had zero bypass cases.

**F2 (High) — shell injection via server-controlled storageKeyId: FIXED.** Every interpolating `execSync` in secure-storage.ts (keyctl/keychain/dpapi) is now `execFileSync` with an argument array — no shell parsing of any value. Added a `ValidatingStorage` wrapper enforcing a strict `^[A-Za-z0-9:_-]{1,200}$` allowlist at the single choke-point (`getSecureStorage`), so it covers every backend and every caller, plus the same regex on `RemoteConfigSchema.storageKeyId`. Confirmed the guard sits before the sink at enable time.

**F3 (Med) — ops-executor SIGKILL never fired: FIXED, and you nailed it.** The unref'd timer + immediate reject meant `process.exit()` preempted the SIGKILL and orphaned the QEMU grandchild. Now: `detached` process group + group-wide SIGTERM→SIGKILL escalation, settle only after the child `close`s (kill completes before exit), hard cap so it can't hang. The test was rewritten to drive the REAL `runOpsCommand` against a fake renet that spawns a SIGTERM-ignoring grandchild and asserts the GRANDCHILD actually dies (passes, ~1.6s) — the exact assertion the old inline-reimplementation test lacked.

**F4 (Med) — review-gate lost its fallback guards + fence-truncation: FIXED.** Restored graceful `|| fallback` on all five paginated `gh api` reads (transient failure degrades instead of aborting under set -e, so post-findings no longer skips --mark). Findings extraction is now awk-anchored to the block's own last fence, so an embedded code fence in a finding body can't truncate it. **BONUS — your out-of-scope find:** `last_marker_sha` ran `tail -n 1` before the SHA `sed` on a multi-line marker body, so the SHA was always empty and the whole review-dedup was dead (every green push re-reviewed). Fixed: sed before tail.

**On the disclosed incident** (a subagent `rm -rf`'d the runner's pre-existing `.review-scripts/`): thank you for surfacing it per convention. That was the ephemeral CI runner's bootstrap copy, not local work — and it's the same class as the ENOENT that broke inline-posting last run, already fixed in 735a53f22 by staging the review scripts into $RUNNER_TEMP (outside the workspace) so neither a checkout nor a subagent can touch them.

This reply also satisfies the report-reply gate this wave added.

mfbayraktar:
Re: review report 5070673985 — the one finding FIXED in commit 16865258f (verified locally: harness 51/51, shell gates, gate-test wrapper green).

**Finding — command-scan.sh wrapper regex bypass via flagged shells: FIXED, and confirmed real before fixing.** `bash -lc`/`sh -eu -c`/`bash -eux -c` all bypassed the F1 fix because the shell name and `-c` had to be immediately adjacent, and the prose-strip step separately erased the same quoted payload — so the guardrail hooks exited 0 while the shell genuinely executed. Reproduced live before fixing (`block-admin-merge.sh` returned exit 0 on the `bash -lc` form), then fixed by tolerating any number of flag tokens (bundled or separate) before the final `-c`-ending token on both the extraction and prefix-strip regexes — they have to match the same shape or the payload isn't stripped to command position. Added 3 adversarial cases (bash -lc, sh -eu -c, bash -eux -c) to the harness; reproduced the exact bypass blocking afterward.

**Verified fixes (F2/F3/F4) confirmed correct, no action needed** — thank you for the careful re-read on each, especially catching that the F3 test genuinely drives the real function.

**Process note, not yours to fix:** this report has no `json:review-findings` fence, so `--post-findings` correctly found nothing to post (0 inline comments) despite a real High-severity finding in the prose. That's a gap in the review pipeline's own reliability, not this report — flagging it for a follow-up look at why the follow-up prompt's mandated fence didn't get emitted this pass, separate from the finding itself which I've verified and fixed above.

mfbayraktar:
Re: review report 5071673291 — the finding FIXED in commit bd7eabcd4 (verified: harness 54/54, shell gates, gate-test wrapper green).

**Finding — GNU long options / value-taking flags before -c: FIXED, and you were right that this pattern would keep recurring.** Your own live `grep -oE` runs against `bash --posix -c`, `bash --norc -c`, and `bash -o pipefail -c` were the decisive evidence — reproduced with the exact WRAPPER_RE and confirmed zero match before touching anything.

Per your own suggested fix direction, I stopped enumerating flag shapes entirely and switched to token-scanning: an awk pass that treats ANY intervening token between the shell name and `-c` as skippable, and only asks a closed question — "is THIS token the -c selector" (exactly `-c`, or a single-dash bundle ending in `c`). No flag syntax needs to be recognized at all, so this class of finding (a fourth flag shape, a fifth, ...) can't recur the same way. This also removed the older design's structural flaw you didn't call out but that caused all three rounds: an extraction regex and a separate prefix-strip regex had to independently agree on the exact same shape — that duplication was itself the bug source. The new function emits the payload directly; nothing left to keep in sync.

Verified with benign placeholder commands (echo instead of the real banned string) so verification itself never re-triggers the hook. Added your 3 suggested adversarial cases; harness now 54/54 (was 51, was 48).

mfbayraktar:
Re: review report 5073133936 — the finding FIXED in commit bc2b0890e (verified: harness 56/56, shell gates, gate-test wrapper green).

**Finding — path-qualified shell name bypass: FIXED, and your framing of the pattern is exactly right.** `/bin/bash -c`, `./bash -c`, and `/usr/bin/bash -c` all bypassed the token-scanner because the shell-name test required an exact bare match — same enumeration trap the flag side already escaped, just moved to the shell-name side. Reproduced live before fixing (the exact anchor failed a `grep -x` probe against a path-qualified token, matching your isolated verification method for this sandboxed environment).

Fixed per your own suggested direction: match the token's basename (strip a leading path prefix on a copy used only for the name test) rather than enumerating literal path prefixes as more alternatives — which would just repeat the trap one more time. Verified `/bin/bash`, `./bash`, `/usr/bin/bash`, bare `bash`, and `/usr/bin/env bash` all extract correctly with benign placeholder commands (not the real banned string, which the fixed hook now correctly intercepts), and a plain script invocation (`bash script.sh --flag`) still correctly extracts nothing. Added 2 adversarial cases; harness now 56/56 (54 → 56).

**GraphQL retry fix (24b03436): confirmed correct, no action needed** — thank you for tracing through the empty-vs-unparseable distinction explicitly; that was the exact gap the round-41 crash exposed.

**On the disclosed dirty-runner-checkout note:** verified locally — my working tree is clean, matches origin exactly at every commit, and every file you named (worklist.py, settings.json, the three block-*.sh hooks) is present and intact here. That was the ephemeral Actions runner's own checkout state during your review pass, not a defect in this PR's committed history or my local tree.

mfbayraktar:
Re: review report 5073738233 — the finding FIXED in commit d6aec6fa0 (verified: harness 58/58, shell gates, gate-test wrapper green).

**Finding — quoted shell path defeats the basename strip: FIXED.** Your hand-traced character analysis was exact — `sub(/.*\//, "", name)` on `"/bin/bash"` matches `"/bin/` greedily (the last `/` is inside the quotes) and leaves `bash"`, which fails the exact-match test the same way the bare unquoted case did before round 43. Fixed by also stripping quote characters (both kinds) from the name-only copy alongside the path-prefix strip. Verified quoted and unquoted `/bin/bash`, `./bash`, bare `bash`, and `/usr/bin/env bash` all extract correctly with benign placeholders; the plain-script negative control still extracts nothing. Added 2 adversarial cases; harness now 58/58 (56 → 58).

Operator note (not yours to act on): review passes on this PR are being capped at 3 per commit going forward, so this is likely the last exhaustive pass on this specific function for now.

PR #540, opened 2026-07-27
fix(breakpoint): named-mode minimum duration + first CI validation of the breakpoint work

mfbayraktar:
## Landing status

`rediacc/account#68` squash-merged as **`58515ea`**; this branch's `private/account` pointer now references that commit on `account/main` (squash tree verified byte-identical to the branch tip `b9a6d70`).

## What this is

The named-mode minimum-duration guard, plus **the first full CI validation of the Breakpoint work**, plus the `cf-breakpoint` rotation entry.

Everything landed on `main` earlier today went in with `[skip ci]` for fast iteration, so the last Console CI run on `main` was the merge commit `df5ee0350`. Five commits of changes had never been through a gate in CI. This PR runs them.

**Submodule PR: rediacc/account#68** (`private/account` — registers `cf-breakpoint` in `rotation-manifest.json`).

## The fix in this branch

Named mode fronts the box with Cloudflare Access, whose one-time-PIN login is **two email round trips**. Teardown deletes the Access application the instant the timer expires, so a short named session dies mid-login. Observed live on run `30259141278` (duration 5):

```
10:43:47  URL emailed to the operator
10:48:48  Access application DELETED by teardown
```

Cloudflare answered the callback with **"That account does not have access."** — which blames the policy, the one component that was correct, and sent the operator hunting a permissions bug that did not exist. Preflight now refuses named sessions under 15 minutes and names the two ways forward. Quick mode is untouched (no login step), and the test asserts that explicitly so the guard cannot silently widen.

Not done, deliberately: *"give teardown a grace period for an in-flight login."* It is not a fix — when the timer expires the box itself goes away, so letting someone finish authenticating would hand them a dead origin.

## What CI validated here (round 1: 77 green, 0 failed)

The diff is small; the tree it ran against was not.

- **`Breakpoint Lifecycle` — executed for the first time ever, green.** It is gated on `full_suite` (`event != 'push'`), so every push to `main` skips it; confirmed skipped on the merge commit. A PR is the only way it can run.
- **`check:ci-actionlint` ran in CI**: `✓ actionlint clean across 25 workflow file(s)`. New gate, pinned actionlint 1.7.12, covering every workflow **including the vendorable template under `.ci/breakpoint/workflow/`** that `.github`-scoped gates cannot see.
- **44 inline `run:` ShellCheck fixes** across `ct-tests.yml` (16 sites), `ci.yml`, `cd-v2.yml`, `promote-stable.yml`, `housekeeping.yml`, `ci-build-*.yml` — live CI/CD files, now proven to parse and execute in the real pipeline.
- **`test-breakpoint-drift.sh`** (13 assertions) and **`test-breakpoint-secret-exposure.sh`** (7), plus a regenerated drift manifest.

Round 1 self-healed: `Smoke Test Preview` failed on attempt 1 and the watchdog classified it transient and re-ran it green (`run_attempt: 2`).

## Verified live before this PR

quick + named end-to-end, Cloudflare Access login (correct address admitted, wrong address silently refused), `services=onprem`, XFCE desktop, tmate shell, email and logs delivery channels, and teardown on three paths — timer expiry, manual cancel, and identity recomputed from label + run-id. Cloudflare confirmed clean by API after each: 0 tunnels, 0 DNS records, 0 Access apps.

## Also fixed here: the review pipeline was wiping its own scripts

`claude-review-reusable.yml` checks out `console@main` into `.review-scripts`, then checks out the caller repo at the workspace root. That second checkout defaulted to `clean: true`, wiping `.review-scripts`. Two different wrong outcomes:

- **Submodule callers** have no `.ci/scripts/review/`, so the fallback `bash .ci/scripts/review/bootstrap-review-scripts.sh` did not exist and the job died with **exit 127 before any review ran** (rediacc/account#68, run `30264668660`).
- **console-as-caller** "worked" — but only because the bootstrap copied the PR's *own* scripts into `.review-scripts`, defeating the property asserted three lines above it: *"a PR can never substitute its own review logic."* A console PR could have edited the thing that reviews it.

`clean: false` is the whole fix; main's copy now wins unconditionally, and the bootstrap becomes the self-disabling pre-first-merge fallback its header claims to be. Same flag and reason as the "Checkout PR head" step that already carries it.

**Timing constraint, stated plainly:** callers consume the reusable `@main`, so this fix cannot take effect until it lands on main. `rediacc/account#68`'s review check stays red for the rest of this wave and cannot be made green pre-merge. It does not block console — round 2 was fully green with that check already failing.

## Also fixed: `--connect-timeout` was ignored on the quick path

`start-tunnel.sh` called `bp_wait_for_transport "$url" 60` with a hardcoded 60 on the quick path, while the named path passed `"$CONNECT_TIMEOUT"`. The flag is advertised in the usage line and applied only to the URL-publish wait — silently ignored for the transport wait.

It cost a CI round: `Breakpoint Lifecycle` went red on run `30270471731` when trycloudflare's edge did not route within 60s. The breakpoint code was byte-identical to the previous green round (`git diff 396e5c6a9..64541a1f5` lists only the review workflow), A1/A2 passed, and the URL was published — so cloudflared started and this was the edge, not us.

Raising the ceiling cannot cure an edge outage. What it fixes is that a caller asking for more patience now gets it, and the error now says which half failed.

## Still untested after this

- **Force-cancel drill** — the watchdog path that bypasses `always()`. The one that would actually leak.
- **Nightly sweeper** — wired at `housekeeping.yml:80`, never run, not even `--dry-run`.
- Gateway `/desktop` `/vnc` `/health` routes (only the catch-all is proven), non-default runners, `hold-on-failure`, concurrent sessions, `vm-os` (deliberately unimplemented).


mfbayraktar:
Reply to review report 5093471383 (https://github.com/rediacc/console/pull/540#issuecomment-5093471383).

Verdict was **no blocking correctness/security/performance defects**, and I am not treating that as "nothing to say" — here is what the review covered and why no code changed in response:

- **`preflight-breakpoint.sh` minimum-duration guard** — accepted as-is. The 15-minute floor is empirical, not a guess: run `30259141278` emailed the URL at `10:43:47` and teardown deleted the Access application at `10:48:48`, so the operator had ~5 minutes to complete a login that is two email round trips. Quick mode is deliberately exempt (no login step) and `test_named_refuses_too_short_a_duration` asserts that exemption so the guard cannot silently widen.
- **`start-tunnel.sh` connect-timeout fix** — accepted. Both call sites now pass `"$CONNECT_TIMEOUT"`; the quick path previously hardcoded `60` while advertising the flag. Stated limitation, unchanged by the review: I could not build a deterministic repro of a transport timeout, because forcing Cloudflare's edge to publish a URL and then not route it is not controllable from this side. What is verified is that the flag reaches the function and that the two call sites are now textually identical.
- **`claude-review-reusable.yml` `clean: false`** — accepted, and it is the highest-risk change here. It stops the caller checkout wiping `.review-scripts`. The half that matters is not the exit-127 on submodule callers but that console-as-caller was silently falling through to the bootstrap and **reviewing with the PR's own scripts** — the property the surrounding comment claims is impossible. Note this review itself ran main's copy, so it could not exercise the fix.
- **`test-breakpoint-mode-selection.sh` / MANIFEST / gitlink** — accepted. The manifest is generated (`check-breakpoint-drift.sh --write`), never hand-edited.

No changes made in response to this report. Known-red and deliberately not fixed here: `rediacc/account#68`'s own review check, which cannot pass until this PR lands on `main` (callers consume the reusable workflow `@main`).

PR #542, opened 2026-07-28
chore(submodules): point private/account at the squashed dep bump

mfbayraktar:
## What

One line: `private/account` moves from `b0ea51f` to `bfc9daea`.

`rediacc/account#69` (@hono/node-server 2.0.11 -> 2.0.12) squash-merged after
console#541 landed. `main` was left pinned to the pre-squash branch commit,
which stays reachable only while branch `0727-2` exists in that repo. This
re-points it at account's `main`.

## Why it is a separate PR

Recorded during Wave A. console#541 had to merge first, because account's
Claude Review resolves its tooling from `console@main`; reviewing account#69
any earlier meant reviewing it with the very tooling console#541 was fixing.
That ordering costs exactly one extra pointer commit, which is this one.

## Related

- rediacc/account#69 (merged as `bfc9daea`)
- Follows console#541

## Testing

No source change, so there is nothing to unit test. The relevant assertion is
`Submodule Branches`, whose rule 1a passes a pointer bump automatically when
the commit is an ancestor of the submodule's `origin/main`. `bfc9daea` is
account's `main`, so this PR is the intended shape for that rule rather than
an exception to it.

## Note for the CI overhaul program

This PR is a clean natural specimen of the case `pointer_bump_only` was built
to shortcut, and it will NOT take the fast path: `detect-pointer-bump.sh` runs
on `pull_request`, where `actions/checkout` hands it the synthetic two-parent
`refs/pull/N/merge` commit that its own walk aborts on. That fast path has been
false on 12 of 12 sampled runs. Its cost here is the "before" measurement for
Wave B.


mfbayraktar:
Closing as redundant: #543 already contains this commit in full (`3e1b6240c` is in its history, and both branches carry the same `bfc9daea` gitlink for `private/account`), so the pointer bump ships with Wave B in a single merge.

Opening it separately was my mistake. I created it before deciding Wave B would branch from it; once `0728-2` was branched off `0728-1` this became a strict subset, and I did not revisit it. The result was two concurrent full CI rounds (~500 machine-minutes each) for one change. Its run has been cancelled.

The underlying reason a superseded run was not cancelled automatically is fixed in #543: `ci.yml`'s concurrency group keyed PR runs on `github.run_id`, which is unique per run, so every PR run sat in a group of one and could never supersede its predecessor.

PR #553, opened 2026-08-07
fix(cd): every version check that could not fail — the 1.2.16-shipped-as-1.2.17 class

mfbayraktar:
On 2026-08-07 a release published CLI binaries built as **1.2.16** under the label **1.2.17**. Nothing user-facing broke, because `tag-and-release` needs `validate-install-published` and so was skipped — the channel pointers never advanced. But three separate guards should have caught it and **every one was incapable of failing**.

A read-only sweep then found ten more instances of the same class. The operator's rule for the fix:

> **WE SHOULD ALWAYS HAVE A VERSION! SKIP OR FAIL!**

Every path must end verified, or in an explicit visible skip, or in a failure. A path that cannot determine a version and returns success **is** the defect.

## The guards that could not fail

**`assert-artifact-version.sh` had never once run in its life.** It downloads an artifact named `cli-manifest`; nothing in the repo ever produced one — the manifest existed only inside the ~7 GB `release-artifacts-<sha>`. So it took its not-found → warn → `exit 0` branch on *every release since it was written*. `cd-stage.yml` now uploads the manifest alone under that name with `if-no-files-found: error`, and all three fail-open exits are hard failures. (Already on main; the first release after it printed `Artifact version v1.2.17 matches promotion target v1.2.17` — a line that had never existed.)

**`verify_version()`**, which both other guards call, was `echo "$output" | grep -q "$expected"`. That passes on an **empty expectation** (`grep -q ""` matches anything), on **empty output**, and on a **substring** — so `1.2.1` "verified" against `1.2.16`, and the dots were regex wildcards. Now exact token matching, failing loudly on all three.

**Windows install validation** ran `powershell --version` and discarded the output, so both Windows platforms reported PASS on the very artifact Linux and macOS failed.

## The ten found by the sweep

- **Seven of eleven install methods never compared a version at all.** `$VERSION` appeared zero times in those functions, though CI passes `--version` into every one. Fixed with a container-side output fence, deliberately *not* a host-side grep: apt, npm and brew each print the expected version while installing, so matching the transcript would accept `Setting up rediacc-cli (1.2.17)` while the binary reports 1.2.16 — a check that cannot fail, rebuilt.
- **The test script exited 0 after running zero tests.** A typo'd `--method` was swallowed and success meant "nothing failed".
- **Retry-mode releases skipped the artifact assertion**, justified by a comment claiming the artifacts "already match by definition". They do not: retry takes its version from the current tag but its artifacts from the latest green CI on main.
- **The build's own smoke test compared nothing** — non-empty and not-`null`, while clobbering the variable holding what it was told to build. It is the only place that reads a version out of real bytes.
- **`inject-env.sh --strict`, the only `0.0.0-dev` guard, had zero callers.** Now wired at four release-path boundaries, gated to push-to-main; dev, PR and fork builds keep the fallback.
- **Attestation verification had no failing exit path**, and its `find` over two absent directories exited 0 having verified zero artifacts.
- **A swallowed tag fetch** could silently regress the version — now retried then hard-failed, with the token redacted rather than the stream suppressed.
- **An empty version collapsed the image closure key**, letting a cached image built at an older version be promoted. Now `untagged-<sha>`, distinguishing.
- **`compareVersions` returned "equal" for malformed versions.** Note it is *not* the empty case — `Number("")` is 0 — it is non-numeric segments producing NaN.
- Duplicated unanchored-grep idiom, normalised.

## Verification

Battery **81 passed / 0 failed / 889 assertions**, up from 71/0/820. Ten new gate tests, each with a planted defect **watched to pass before the fix made it fail** — including replaying the incident's own 1.2.17-versus-1.2.16 through the extracted comparison block. `check-ci-parity` green both directions: 192 gates, 81 battery tests.

## Regression found and fixed within this PR

Hardening `compareVersions` to throw exposed a real hazard: `private/account/src/app.ts:548` awaits `checkCliVersionPolicy` with no `try/catch`, feeding it the client-controlled `x-cli-version` header. A malformed header would have gone from **silently bypassing** the minimum-version gate to a **500 on attacker-controlled input**. Both wrong.

Fixed in **[rediacc/account#75](https://github.com/rediacc/account/pull/75)**, pointer-bumped here: a malformed version is now **rejected** with the same 426 the out-of-date path returns, `code: CLI_VERSION_INVALID`. The guard sits *before* `compareVersions`, so the throw is unreachable rather than merely caught. Verified against the real primitive — `garbage`, `1.2.x`, `<script>` all rejected before it runs, `1.2.17` compares normally.

Two related gaps closed in the same pass:

- `isValidVersion` and `InvalidVersionError` existed in `version.ts` but were never re-exported from the utils index, so the only thing consumers could import was the function that throws. **A guard callers cannot reach is a guard that cannot be used** — the same shape as `--strict` with no callers, and as the `cli-manifest` guard looking for an artifact nobody produced. Three instances of one pattern in this wave.
- `checkForUpdate`'s bare `catch {}` now logs via `REDIACC_DEBUG`. Silent to the user was right and stays; silent to us meant a malformed manifest went from "returns 0, reports up to date forever" to "throws, reports up to date forever" — same outcome, still undiagnosable.

Draft until CI reports.

## Submodule PR

- `private/renet` → **rediacc/renet#99** (branch `0807-2`) — `klauspost/compress` v1.19.1 → v1.19.2, which is what Console's `Quality / Go` freshness check demanded.
- `private/account` → **rediacc/account#77** (branch `0807-2`, same name as this one)

That PR rejects a malformed `x-cli-version` instead of bypassing the minimum-version gate or 500ing on it, and it needs the `isValidVersion` export this PR adds. Land this one first.

**Last pushed:** `c60def0d6`

- `c60def0d6` Merge remote-tracking branch 'origin/main' into 0807-2
- `1ebd8aff4` fix(review): address all three review findings on #554
- `7fa0ce1ab` test(review): pin the cap reached by SPENT attempts, not posted reports
- `927793a6b` docs(ci-overhaul): the deadlock guard, the turn budget, and why a retry lies
- `bcd40d9c8` fix(review): the deadlock guard could not fire, so a capped PR went unmergeable


mfbayraktar:
Superseded by #555, which is **MERGED** as `753a2f48f`. Identical content, linear history.

This PR was unmergeable for a structural reason, not a defect in its content, and it is worth recording precisely:

1. `main` moved (#552 and #554 landed), so this branch went `DIRTY`.
2. Updating it by **rebase** would have required a force-push. `block-git-force-push.sh` is categorical: *"Rewriting already-pushed history is the user's decision, not an agent's."*
3. So it absorbed `main` via a **merge commit** — a plain fast-forward push, nothing rewritten.
4. This repo is **rebase-merge only** (verified, not assumed: `allow_merge_commit=false allow_squash_merge=false allow_rebase_merge=true`), and GitHub answered `This branch can't be rebased`.
5. The base-flip workaround failed too: `0807-1` had been deleted on merge, so there was nothing to flip to.

The rebase performed before step 3 was preserved and pushed as `0807-4` — **17 commits, zero merge commits** — which rebase-merge accepts. Its tree is byte-identical to the one that passed **79 jobs green** here, plus a `knip 6.31.0 -> 6.32.0` bump for a freshness gate that fired mid-flight.

Nothing from this PR is lost; all of it is on `main`.

PR #557, opened 2026-08-08
feat(ci): gate-id convention, auto-labels from the review pass, 18-key greenlight, renet pointer (big-bang)

mfbayraktar:
## What (big-bang, operator-directed)

Four workstreams on one branch, one CI round:

### 1. `check:ci-gate-id-convention` (the original PR)

A convention held by 57 gate scripts and enforced by none: a reviewer caught the
58th registered under a `check:ci-*` alias and correctly called it "not a defect,
just an inconsistency" -- which is exactly what the next person copies. Now
enforced, narrowly: only what a manifest entry's `run` executes (one npm-alias hop
included), never the `ci.test` coverage field. Control plants the exact shape that
shipped; FLOOR of 40 parsed entries.

### 2. Auto-labels from the review pass that already runs

**The bump labels were dead.** `detect-bump-type.sh` grepped `(#N)` from HEAD
commit titles -- squash-merge output; this repo went rebase-merge 2026-07-30 and 0
of the last 60 main commits carry it, so every release fell back to patch
regardless of labels. Rewritten: `<latest tag>..HEAD` resolved through
`commits/<sha>/pulls` (follows rebased commits), highest-priority label across the
union of merged PRs.

The labels ride the existing Claude review invocation -- a second `json:pr-labels`
fence in the same report, zero extra AI calls. Mechanical path floor (docs/ci)
lands even when the review starves; hard whitelist between model and API;
`bump-major` absent from it by construction (major = recommendation-only); ledger
reconciliation so hand-applied labels survive. Review scripts execute from main,
so the applier arm is a guarded no-op until this merges; the two mutation-proven
gate tests are the pre-merge evidence. Also fixes a `--mark` honesty hole (the
pipeline's own ledger comment counted as review output) and its spent-attempt
sibling.

### 3. Cross-PR greenlight: 2 keys -> 18

The skip-if-identical-inputs-already-green mechanism existed and worked (it set
`run_renet=false` on a live run) but only 2 of 18 keys were enrolled, so a
quality-script/docs PR still paid the full VM battery. Now all 18: `jobNames[]`
for matrix legs (all five e2e_workers legs must be green; one missing refuses),
`submodules[]` with empty legal, shared VM_E2E closure for the eight VM legs,
walk budget 60->90s (live 18-key walk: 16.6s). Two-layer wrong-skip defense; the
new closure-trace gate caught a real hole in the live renet closure
(`app-token` missing) on its first run.

### 4. Riders

[...]

PR #558, opened 2026-08-08
fix(hooks)+perf(ci): Stop-hook task-queue visibility; W/S/T-parallel quality-gate battery (3.14x)

mfbayraktar:
## What

Two proven workstreams from today's operator directives:

### 1. The Stop hook could not see the harness task queue (two stacked bugs)

Found when the operator asked why the hook never pinged about a fully planned,
unblocked pending task during an hours-long background wait:

- **The task store moved** to `session-<team-id>` while Stop events carry the
  conversation id, so `pending_tasks()` read a nonexistent directory and returned
  `[]` all session -- the exact "two queues, one supervisor, watching the empty
  one" failure its own v5 docstring describes, reintroduced one naming scheme
  later. Fix: a transcript-content join (TaskCreate result lines carry id +
  subject; the unique directory holding that task is this session's), refusing on
  ambiguity. An EMPTY primary dir defers to the join -- the live proof found a
  zero-file July relic shadowing the real store.
- **Pure-background-wait never consulted the queue** even when visible. Now an
  unblocked pending task makes the wait impure: the check-in fires as WORKABLE
  TASKS and names it; the honest silencer is a recorded blocker.

Proof: 7 unit cases, live resolve of the real store, and the check FIRING IN
PRODUCTION on its first eligible stop, naming the exact task that motivated it.
Harness 589/0 (new case pair 163y: fire + blocked-control).

### 2. Quality-gate battery: 18 min serial -> ~5 min, inside the same step

87 tests ran serially, 18 of the Security job's 20 budget minutes. A flat pool is
unsafe for a specific reason: two tests write fixtures INTO THE REAL TREE while
~19 others recursively enumerate those directories (the observed flake shape).
W/S/T schedule instead: writers as one serial chain, temp-isolated tests fill
slots, real-tree readers released on the writers' done-marker. Workers never
print; main prints blocks in glob order, so the transcript is the serial
transcript. Missing result = FAILURE, never a skip.

Measured, same code both ways: **956s serial vs 304s parallel (3.14x)**,
identical failure sets and assertion totals. `gate-test:run-all-parallel` pins
overlap (with a must-be-slow serial control), byte-for-byte jobs=1/jobs=4
agreement, non-interleaving, and the W/S hold-back -- mutation-proven at TWO
seams (wait loop deleted; done-marker falsified), both RED.

Rider: `log_info`/`log_error` were called by a gate test but defined nowhere,
silent only because that file runs without `-e`. Defined centrally.

## Notes for review

- This is the first PR through the fully-live auto-labels path (fence + applier
  landed with #557), and the second live acceptance of 18-key greenlight: the
  diff touches no closure input, so the whole VM battery should skip against the
  75c4de9af evidence run.
- Branch cut from pre-release main; will pick up the v1.2.21 release-state
  commits via update-branch/rebase as the operator noted.

**Last pushed:** `a077bd32f`

- `a077bd32f` docs(agent): triage notes for the parallel quality-gate battery
- `58c86f081` perf(ci): parallelize the quality-gate battery inside its step (W/S/T)
- `af5fb19ee` fix(hooks): the Stop hook could not see the harness task queue, twice over
- `b652e4d5a` chore(submodules): renet pointer to merged main (renet #102 rebased tip)
- `198b52050` docs(ci): record the accepted one-hop alias limit in the gate header


mfbayraktar:
Superseded by #559: this branch was cut from pre-rebase 0808-2 SHAs, and after #557's rebase-merge plus the v1.2.21 release commits the PR became CONFLICTING (no test-merge commit is buildable, which is why GitHub created zero pull_request runs for it -- both the opened and synchronize events produced nothing). Force-push is hook-blocked by policy, so the same three commits were cherry-picked onto post-release main as 0808-4. Content identical, proofs unchanged.

PR #559, opened 2026-08-08
fix(hooks)+perf(ci): Stop-hook task-queue visibility; W/S/T-parallel quality-gate battery (3.14x)

mfbayraktar:
## What

Two proven workstreams from today's operator directives:

### 1. The Stop hook could not see the harness task queue (two stacked bugs)

Found when the operator asked why the hook never pinged about a fully planned,
unblocked pending task during an hours-long background wait:

- **The task store moved** to `session-<team-id>` while Stop events carry the
  conversation id, so `pending_tasks()` read a nonexistent directory and returned
  `[]` all session -- the exact "two queues, one supervisor, watching the empty
  one" failure its own v5 docstring describes, reintroduced one naming scheme
  later. Fix: a transcript-content join (TaskCreate result lines carry id +
  subject; the unique directory holding that task is this session's), refusing on
  ambiguity. An EMPTY primary dir defers to the join -- the live proof found a
  zero-file July relic shadowing the real store.
- **Pure-background-wait never consulted the queue** even when visible. Now an
  unblocked pending task makes the wait impure: the check-in fires as WORKABLE
  TASKS and names it; the honest silencer is a recorded blocker.

Proof: 7 unit cases, live resolve of the real store, and the check FIRING IN
PRODUCTION on its first eligible stop, naming the exact task that motivated it.
Harness 589/0 (new case pair 163y: fire + blocked-control).

### 2. Quality-gate battery: 18 min serial -> ~5 min, inside the same step

87 tests ran serially, 18 of the Security job's 20 budget minutes. A flat pool is
unsafe for a specific reason: two tests write fixtures INTO THE REAL TREE while
~19 others recursively enumerate those directories (the observed flake shape).
W/S/T schedule instead: writers as one serial chain, temp-isolated tests fill
slots, real-tree readers released on the writers' done-marker. Workers never
print; main prints blocks in glob order, so the transcript is the serial
transcript. Missing result = FAILURE, never a skip.

Measured, same code both ways: **956s serial vs 304s parallel (3.14x)**,
identical failure sets and assertion totals. `gate-test:run-all-parallel` pins
overlap (with a must-be-slow serial control), byte-for-byte jobs=1/jobs=4
agreement, non-interleaving, and the W/S hold-back -- mutation-proven at TWO
seams (wait loop deleted; done-marker falsified), both RED.

Rider: `log_info`/`log_error` were called by a gate test but defined nowhere,
silent only because that file runs without `-e`. Defined centrally. Plus triage
notes in docs/agent/ci-gates.md for future battery-failure debugging.

## Provenance

Replaces #558 (branch cut from pre-rebase SHAs went CONFLICTING after the
v1.2.21 release commits; force-push is hook-blocked, so the same three commits
were cherry-picked onto post-release main). First PR through the fully-live
auto-labels path; second live acceptance of 18-key greenlight (diff touches no
closure input -- the VM battery should skip against the 75c4de9af evidence run).

**Last pushed:** `a9b438b5a`

- `a9b438b5a` fix(hooks,review): answer the #559 review; sweep the task-dir fix, commit its coverage, and fix the label applier's first live defect
- `75c4cb155` docs(agent): triage notes for the parallel quality-gate battery
- `05fe8d5fb` perf(ci): parallelize the quality-gate battery inside its step (W/S/T)
- `998bb7b91` fix(hooks): the Stop hook could not see the harness task queue, twice over
- `b8e332b73` chore(release-state): advance contract floor to v1.2.2 [skip ci]

### DIFF
--- a/.ci/scripts/review/claude-review-gate.sh
+++ b/.ci/scripts/review/claude-review-gate.sh
@@ -380,6 +380,31 @@ if [[ "${1:-}" == "--apply-labels" ]]; then
         ' <<<"$report") || verdict=""
     fi
 
+    # FALLBACK: the fence may live only in the POSTED COMMENT. Observed on the
+    # feature's first live run (#559, run 31267699743): the model posts its
+    # summary itself via `gh pr comment` and put the fence THERE, while its
+    # final result text back to the harness did not repeat it -- so the
+    # extraction above logged "no json:pr-labels block" beside a PR whose
+    # verdict comment plainly carried one. The result text stays the primary
+    # source (it wins when both exist); the newest fence-bearing comment is
+    # the fallback. Trust model: PR comments already carry the reviewed-SHA
+    # markers this pipeline acts on, the whitelist hard-filters every label,
+    # and bump-major is never applied automatically, so a forged fence cannot
+    # do more than a forged marker could.
+    if [[ -z "$verdict" ]]; then
+        comment_report=$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" --paginate \
+            --jq '.[] | select(.body | contains("```json:pr-labels")) | .body' 2>/dev/null | tail -c 65536) || comment_report=""
+        if [[ -n "$comment_report" ]]; then
+            verdict=$(awk '
+                /^[[:space:]]*```json:pr-labels[[:space:]]*$/ { capturing = 1; n = 0; next }
+                capturing && /^[[:space:]]*```[[:space:]]*$/ { capturing = 0; next }
+                capturing { buf[++n] = $0 }
+                END { for (i = 1; i <= n; i++) print buf[i] }
+            ' <<<"$comment_report") || verdict=""
+            [[ -n "$verdict" ]] && log_info "json:pr-labels fence found in the posted report comment (absent from the result text)"
+        fi
+    fi
+
     if [[ -z "$verdict" ]]; then
         log_info "no json:pr-labels block in the report; mechanical labels only"
     elif ! jq -e '
--- a/.ci/scripts/test/gates/test-review-labels.sh
+++ b/.ci/scripts/test/gates/test-review-labels.sh
@@ -623,6 +623,40 @@ test_mark_does_not_count_the_ledger_comment_as_output() {
 # ---------------------------------------------------------------------------
 # Advisory end to end
 # ---------------------------------------------------------------------------
+test_fence_only_in_posted_comment_is_found() {
+    local t="$1"
+    setup "$t"
+    # The feature's FIRST LIVE RUN (#559, run 31267699743): the model posted its
+    # summary itself and put the fence in the COMMENT; its result text back to
+    # the harness did not repeat it. The old extraction read only the result
+    # text, logged "no json:pr-labels block", and applied nothing beside a PR
+    # whose verdict comment plainly carried a verdict.
+    execution_file "$t" "$(report_with_verdict '')"
+    comments_fixture "$t" "$(jq -n --arg body "## Verdict: request changes
+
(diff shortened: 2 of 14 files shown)


mfbayraktar:
Re: review summary 5227094422 -- both findings accepted and fixed in `a9b438b5a`.

**Finding 1 (high, the unswept call sites): fixed, and the sweep went further than the suggested fix direction.** Threading `transcript_path` into `wl_liveness.py` `ladder()` and both `wl_store` signature functions is done as you suggested (all three now resolve through the join). But the fourth caller, `worklist.py:884` (the `--state` CLI verb), has NO transcript in scope -- and threading alone would therefore have left the CLI writer and the hook checker hashing DIFFERENT task sets whenever the primary dir is empty, making STATE.md eternally stale (writer stores a sig over zero tasks; checker computes one over the real set). The fix persists the resolution once found (`/tmp/.../claude-worklist/taskdir.<sid8>`, validated-before-trust: the cached dir must still exist and still hold a task, tmp+rename write) so every transcript-less process agrees with the battery. The CLI call site is deliberately unchanged; it rides the durable cache.

**Finding 2 (medium, the uncovered core fix): you read the diff correctly -- the claimed "7 unit cases" ran as ephemeral scratchpad python during development and never landed in the PR.** The description overstated what the diff carried. Now committed as harness case 163x, which is stronger than the original scratchpad cases because it drives the whole chain through the REAL hook: primary `session-deadbeef` EMPTY, tasks planted under a differently-named `session-*` dir, resolution off a `Task #21 created successfully: ...` line in the synthetic transcript, asserted end-to-end by the workable-tasks check-in NAMING the task. Its control plants TWO matching dirs and asserts the join refuses: no task manufactured, the wait stays pure. Harness total: 591 passed, 0 failed.

**Nits**: manifest audit comment re-indented to its block. The 6000ms ceiling in `test_pool_runs_tests_concurrently` stays as-is deliberately -- it is 3x the theoretical 2s floor, and its paired jobs=1 control (>= 8s) keeps the measurement falsifiable; if it ever flakes on a contended runner the control tells us whether the stopwatch or the pool is lying.

**Coverage map**: nothing flagged unreviewed in your summary; nothing to rule on there.

**One thing your pass surfaced beyond its findings, fixed in the same commit**: your own `json:pr-labels` fence (`{"bump":"patch","kind":["bug","ci"]}`) was emitted correctly but landed only in your posted comment, not in the result text the applier reads -- so the Apply-labels step logged "no json:pr-labels block" and the ledger recorded an empty apply. The posted-comment fence is now the fallback (result text stays primary), with two new gate cases pinning the live shape and the priority order. Like every review-pipeline change this is inert until main; the first PR reviewed after this merges gets its labels applied.

mfbayraktar:
Re: review report 5227310677 and follow-up summary 5227309956 -- acknowledged in full.

**Findings**: none posted (`review-findings: []`); the two prior blocking findings verified resolved by this pass, including the durable-cache isolation story in the harness (the TMPDIR-scoped wipe the review traced is exactly the mechanism the CONTROL case relies on).

**Coverage**: the delta read covered every changed file including the self-reported applier fallback, with the priority order and newest-fence resolution verified in the awk itself -- nothing flagged unreviewed, nothing to rule on.

**One correction to the delta note, for the record**: the 6000ms-ceiling nit was left alone deliberately (stated in the prior answer), not merely because the file was absent from the delta -- both readings agree on the outcome.

Nothing outstanding. Proceeding to merge; the label fence emitted by this very review lands via the now-merged fallback starting with the next reviewed PR.

mfbayraktar:
review-status nudge: head a9b438b5a is reviewed (marker matches, follow-up verdict: nothing blocking, report 5227310677 answered as 5227315377); requesting currency evaluation.

PR #563, opened 2026-08-09
feat(hooks): enforce /handoff deliverables and waves from the Stop hook

mfbayraktar:
## What this does

`/handoff` produced prose and nothing else. Its scope waves lived in a README paragraph, its deliverables lived in the command's own step list, and the only thing that ever asked the consuming session to seed the worklist was an instruction sentence inside `PROMPT.md`. The Stop hook, 13k lines across 13 modules, had **zero** grep hits for "handoff": nothing verified that a producing session actually wrote every deliverable, and nothing verified that the consuming session claimed or finished a single wave. An ignored `PROMPT.md`, or one compacted out of context, silently dropped program work with no red anywhere.

`docs/<slug>/CHECKLIST.md` is now the artifact both ends are held to, and the hook reads it.

```
# Handoff checklist: <slug>
Status: producing            <- first 10 lines; producing|executing|done|superseded
Owner: 99ccf057              <- 8-char session prefix, required while producing

## Deliverables
- [ ] d1 file:docs/<slug>/README.md
- [ ] d2 file:docs/<slug>/PROMPT.md

## Waves
- [ ] w1 Wave A: <one-line title>
```

- **Deliverables are file-verified, period.** The path must exist and be non-empty; `~` expands and relative paths resolve against the repo root. The tick is bookkeeping, the file is the truth, so a ticked-but-missing deliverable is called out loudly rather than believed.
- **Waves are tick-on-trust with store linkage.** A wave is settled by `[x]` or covered by any worklist item carrying the literal token `cl:<slug>/<wN>`, seeded with `worklist.py --add <me> 'cl:<slug>/<wN> <title>'`. Evidence discipline rides the existing `--tick` gate; no new evidence machinery was invented.
- **Lifecycle** is `producing` (blocks its owner until every file verifies and the status is flipped) then `executing` (blocks any stopping session on a wave that is neither ticked nor covered; demands the box tick when the store settled it; demands `Status: done` when everything is settled) then `done` (gated on all-verified plus all-ticked) or `superseded` (terminal).

## Enforcement

Four check keys, riding the existing `vadd()` focused-block rotation:

| key | tier | blocks |
|---|---|---|
| `cl-shape` | ALWAYS when the parser itself crashed, rotated for grammar violations | any stopping session |
| `cl-producing` | rotated | the owner only |
| `cl-flip` | rotated | the owner (untagged owner blocks anyone) |
| `cl-waves` | rotated, one body per slug | any stopping session |

`cl-waves` blocking everyone is deliberate: an uncovered wave is *unclaimed* work, exactly the semantics an untagged worklist item already has. The moment any session adds the `cl:` item the wave has an owner and stops blocking everybody else, so the worst case is one redundant simultaneous block that self-resolves, and every violation has a single-turn solo exit. A foreign `producing` checklist never blocks: it emits the change-latched `cl-foreign` advisory, which gains an adoption hint once the owner's transcript has been idle 24h.

## Cost

The full Stop battery reads checklists because that is the enforcement point: zero checklists costs one glob, and only `producing`/`executing` files are parsed in full. The 5-minute poll fast path stays **read-free by contract**: `clsig` is a sha1 over sorted `(relpath, mtime_ns, size)`, stat-only so a `chmod 000` checklist cannot make it raise, and both `clsig` and `cl_live` are banked in the pollbase. A poll forfeits when the signature moved, when banked `cl_live > 0`, or when the keys are absent from an older baseline. SessionStart and PostCompact inject a LIVE HANDOFF CHECKLISTS listing.

## Surface change vs provably unchanged

- **Changed:** the `/handoff` command now writes `CHECKLIST.md` as its first file and seeds waves with the `cl:` token; the Stop hook gains four blocking checks and one advisory; SessionStart/PostCompact gain one context block.
- **Provably unchanged:** no product code. Nothing under `packages/`, `private/` or `.github/` is touched by this PR. No CLI command added or removed, no user-facing binary surface, no workflow edited. The diff is **eight** files: five under `.claude/hooks/stop/`, one command definition, one design-record doc, and one CI quality gate (`check-python-lint.sh`, added in round 2 below because it could not see this PR's own new module).

## Submodule PRs

**None.** All four submodule pointers were already at their `origin/main` tips when this branch was cut and are unchanged by it: renet `e48cc4ae6`, account `9cde3ea7f`, elite `4f104ab0b`, homebrew-tap `b1e617a4b`. Nothing to bump, nothing rolled back.

## Verification (run, not read)

| gate | command | result |
|---|---|---|
| stop-hook suite | `bash .claude/hooks/stop/test-worklist-v5.sh` | round 1: `passed=626 failed=0` (35 new assertions, cases 193-203). Round 2 supersedes this at 634; see below |
| report-inbox suite | via the CI gate wrapper | `passed=118 failed=0` |
| CI gate wrapper | `bash .ci/scripts/test/gates/test-worklist-hooks.sh` | exit `0`, both harnesses green |
| hook umbrella | `bash .claude/hooks/test-hooks.sh` | `PASS=707 FAIL=0` |
| em-dash scan | `grep -n` over every authored file | clean |

Every new blocking case is paired with a clean-fixture control, including a zero-checklist control that must produce no checklist output at all, and a stat-only unit call against a `chmod 000` file that asserts the fixture is genuinely unreadable before concluding anything from the signature succeeding.

## Live validation, unsolicited

While this branch was being cut, an independent session wrote `docs/backup-storage/CHECKLIST.md` (Status `executing`, 9/9 deliverables ticked and file-verified, 6 waves) and seeded all six `cl:backup-storage/wN` worklist items, exercising both halves of the gate end to end within an hour of it landing. That suite is deliberately **not** in this PR: it belongs to that session.

## Accepted residuals (stated, not solved)

- A session that skips writing a checklist entirely is invisible to the hook. The mitigation is social: the checklist is the first file write and is named in the command's Report and Constraints sections.
- While any checklist anywhere sits at `producing`, every session's poll fast path forfeits and pays the full battery. That follows directly from "polls are never silent while a live checklist exists"; the exit is finishing the deliverables and flipping, or superseding.
- An `executing` checklist whose waves are all ticked blocks immediately demanding the `done` flip, so a quiet satisfied state at `executing` requires at least one covered-but-unticked wave. Case 198b pins this.

Design record: `docs/ci-overhaul/06-progress.md`, dated section. Approved plan: `i-use-handoff-custom-recursive-mist.md`.

---

## Round 2 (7eecc06ff): three defects fixed

The automated review on `71c842d3f` found two real correctness defects in the new enforcement logic. Both were **confirmed against the running system before being fixed**, not accepted on reading. A third was found while fixing them.

**1. Rotation starvation on shared violation keys.** Every violation used a fixed key, and the focused-block picker builds `order = {v[0]: i for ...}` — a dict comprehension, so duplicate keys collapse to one entry, give identical sort tuples, and `min` is stable on the first. Running the picker directly over four stops with two `cl-waves` violations produced `A, C, A, C`: the second checklist was never itemised, only absorbed into the "N more outstanding" count, until the first resolved. Two live handoffs is a real state; this repo had two simultaneously today. The review missed one instance which I found while confirming it: `cl-foreign` has the same bug via `outq_add`, which dedupes by key so the second advisory *replaces* the first. Fixed with one `_ckey(prefix, slug)` helper across all eleven per-checklist sites. The two one-per-stop backstops stay unscoped deliberately: the glob backstop has no slug to scope by and returns immediately, so it has nothing to collide with.

**2. A foreign-owner advisory issuing first-person orders.** When a drifted checklist belongs to another session, the code correctly routed it to the non-blocking advisory but reused `V_CL_FLIP`'s body, which ends "Restore the artifact(s), or write the honest status ... **in this turn**" — an imperative aimed at whoever is reading, about someone else's program. Not theoretical: in this same session a shared `.agent/main/STATE.md` was destroyed exactly that way, a staleness gate told the reading session to rewrite a document and a peer's live campaign notes went with it. New `N_CL_FOREIGN_DRIFT` carries the same rows, names the owner, warns against editing another session's files, and ends with the `Reported, never blocked on.` that the producing-foreign path already used. `V_CL_FLIP` is unchanged on the owner path, where the imperative is correct.

**3. The lint gate could not see this PR's own module.** `check-python-lint.sh` enumerated with `git ls-files`, so an untracked `.py` beside tracked ones was silently omitted. It reported "27 files, All checks passed!" while never reading `wl_checklist.py`, the newest and second-largest module in the Stop-hook program, and a real format violation in it surfaced only from running ruff by hand. The script's own header argues that a linter which cannot run is a gate that cannot fail; this was that, in its own file. One `enumerate_py()` function now serves both the real list and a new CONTROL 3 that plants a runtime-keyed untracked `.py` and asserts the enumerator returns it, so the control cannot drift from what it guards. **27 files becomes 28**, and mutating only the enumerator line makes the control fire with exit 1.

### Verification, re-run after the final edit

| gate | result |
|---|---|
| `bash .claude/hooks/stop/test-worklist-v5.sh` | `passed=634 failed=0` (up from 626; 8 new assertions) |
| `bash .ci/scripts/test/gates/test-worklist-hooks.sh` | `PASS: all 2 stop-hook harnesses green`, report-inbox `118/0` |
| `bash .ci/scripts/quality/check-python-lint.sh` | 28 files, lint and format clean |
| enumerator mutation | `CONTROL FAILED`, exit 1, as designed |

Cases 204-206 were each **observed failing before the fix and passing after**. Case 205 was rebuilt after its first version failed for the wrong reason: draining one advisory latches it in the output queue's `shown` ledger, which suppressed the sibling through the refresh window — correct behavior wearing the bug's costume.

## Land prep (e5e367f4b): rebased onto main, and the time-based deps gate satisfied

Two things happened on the way to landing, neither of them about this PR's content.

**Rebased onto `main`**, which had gained two release commits (`chore(release): update homebrew-tap submodule pointer` and `chore(release-state): advance contract floor to v1.2.6`). That is what cleared the `BEHIND` state; it also invalidated the review marker, which is the ordinary cost of a rebase.

**`check:deps` demanded two bumps** — `@anthropic-ai/claude-agent-sdk 0.3.220 → 0.3.226` and `syncpack 15.3.2 → 15.3.3`. Neither relates to anything this branch changes. The gate is time-based, so it begins demanding a bump the moment a release ages past its freshness window, and every open branch inherits the demand. This is the oldest branch in the stack, so it was failing on both.

Worth recording because it argues against long-lived stacks generally: `syncpack` entered the demand *after* a branch further up had already gone green on the agent-sdk bump alone. A stack that outlives a UTC rollover does not converge on green, it accumulates new reasons to be red.

The lockfile was regenerated with `npx npm@10` rather than the system `npm 11`, so it keeps the canonical form and avoids the cosmetic dev-marker flip CLAUDE.md documents. Verified locally: `check-deps` reports all dependencies up to date (10 blocked, 4 too new).


mfbayraktar:
Replying to the automated review summary: https://github.com/rediacc/console/pull/563#issuecomment-5233099795

Both findings are real. I verified each against the running system rather than accepting them on reading, and both are being fixed on this branch.

### Finding 1 (rotation starvation) — CONFIRMED, and reproduced empirically

Not just plausible: I replicated the focused-block picker exactly, outside the hook, against three violations where two share the `cl-waves` key.

```
violations = [("cl-waves", "SLUG-A"), ("cl-waves", "SLUG-B"), ("cl-shape", "SLUG-C")]
stop 1 shown=['SLUG-A uncovered wave']  n_more=2
stop 2 shown=['SLUG-C malformed']       n_more=2
stop 3 shown=['SLUG-A uncovered wave']  n_more=2
stop 4 shown=['SLUG-B never appears']   <- it does not. A, C, A, C, forever.
```

`order = {v[0]: i for i, v in enumerate(rot)}` collapses the duplicate key, so both entries get identical sort tuples and `min` is stable on the first. SLUG-B is absorbed into "N more outstanding" and never itemized until SLUG-A resolves. Since the entire point of this gate is to name what is missing, a violation that can only ever be counted is close to useless.

The review missed one instance of the same bug, which I found while confirming it: the `cl-foreign` **advisory** has it too, by a different mechanism. `outq_add` dedupes by key, so a second foreign checklist's body *replaces* the first's in the queue rather than queueing beside it. Same root cause, same fix.

Fix: scope every per-checklist key by slug (`cl-waves:<slug>` and so on), keeping the key prefix intact so existing greps on the class still match. The rotation's `served` pruning already drops keys that stop being outstanding, so per-slug keys self-clean.

### Finding 2 (foreign-owner wording) — CONFIRMED, and the risk is not theoretical

Rendered, this is what a session that does **not** own the handoff is currently told:

> A status ahead of its artifacts is exactly how program work drops silently... **Restore the artifact(s), or write the honest status in docs/peer/CHECKLIST.md, in this turn.**

An imperative aimed at whoever happens to be reading, about another session's program. In this very session a shared `.agent/main/STATE.md` was destroyed exactly that way: a staleness gate told the reading session to rewrite a document, the session obeyed, and a peer's live campaign notes went with it (recovered only because the single-slot `.prev` backup was read before the next write clobbered it). So this is a live failure mode in this repo, not a style nit.

Fix: the foreign drift path gets its own constant, carrying the same drift rows but stating the repair belongs to the owning session, ending with the literal `Reported, never blocked on.` that `N_CL_FOREIGN` already uses, and explicitly warning against editing another session's checklist or files. `V_CL_FLIP` keeps its current wording for the owner path, where the imperative is correct.

### Nit (residuals list)

Not taking this one: the rotation case is being **fixed**, not accepted, so listing it under "Accepted residuals" would misdescribe it. The design record gets the per-slug keying rationale instead.

### Tests

Both fixes land control-first. The rotation case must be observed FAILING against the unfixed code before the fix is applied, because a regression test for a starvation bug that cannot starve is worthless. The wording case asserts both directions: foreign owner must produce `Reported, never blocked on.` and must not contain `in this turn`, while the same fixture owned by the running session must still block and still say `in this turn`.

### Coverage-map note

Agreed on scope: no product, CLI or www surface, no submodule pointer changes. Flagging one thing the review marked "skimmed" — the suite claim of 626/626 is independently verifiable and I re-ran it myself rather than trusting the PR text: gate wrapper exit 0 with `passed=626 failed=0` plus `passed=118 failed=0`, and the hook umbrella at `PASS=707 FAIL=0`. The next push will restate those numbers after the fix.

mfbayraktar:
Replying to the follow-up review summary: https://github.com/rediacc/console/pull/563#issuecomment-5233667464

Agreed on the verdict, and thank you for the scope correction, which is worth recording because it nearly misled me too. The three autopilot files (`validate-handoff.cjs`, `test-autopilot-harness.sh`, `autopilot.yml`) are **not** this PR's work. They arrived when GitHub rebased the branch onto main at 19:31:31, which rewrote both commits (`71c842d3f` -> `b55a986b4`, `7eecc06ff` -> `457b9137c`). A `git diff <old-head>..<new-head>` after a rebase mixes the branch's own delta with everything main gained underneath it; `compare/main...457b9137` is the honest scope, and that is the command I will use from here.

Two clarifications on your reading, both confirming it rather than disputing it.

**On the `outq_add` overwrite.** You are right that it was not in the original finding. I found it while confirming finding 1: `outq_add` sets `ekey = key` for non-sticky entries and then matches on that key, so a second foreign checklist's body replaced the first's in place rather than queueing beside it. Same root cause as the rotation collapse, different mechanism, so it is fixed by the same `_ckey` scoping and pinned by case 205. That case was rebuilt once: its first version failed pre-fix for the *wrong* reason, because draining one advisory latches it in the output queue's `shown` ledger and suppressed its sibling through the refresh window. Correct behavior wearing the bug's costume, which would have made a green run mean nothing.

**On the unscoped glob backstop.** Your reading matches the intent exactly. It has no slug to scope by, and it `return`s immediately, so at most one can exist per stop and it has nothing to collide with. The per-file wrapper is scoped, because two unreadable checklists would otherwise evict each other in precisely the way the fix exists to prevent.

**On `check-python-lint.sh`.** Two things you could not see from the diff are worth stating. First, the gap was found by the sub-agent that had just spent a session editing `wl_checklist.py`: the gate reported "27 files, All checks passed!" and had never once read the file being edited, and a real format violation in it surfaced only from running ruff by hand. Second, my first version of CONTROL 3 ran its own copy of the enumeration query, so mutating the real enumerator would have left the control green. I built a check that could not fail inside the commit fixing a check that could not fail, caught it by mutating only the enumerator line, and restructured to the shared `enumerate_py()` you describe. The mutation is the evidence: enumerator-only edit gives `CONTROL FAILED`, exit 1.

Suite is 634/0, up from 626, verified by my own run rather than the worker's report. All three new cases were observed failing before their fix.

mfbayraktar:

PR #561, opened 2026-08-09
feat(ci): full autopilot autonomy incl submodules; profiler persistence + runner-advice gate

mfbayraktar:
## Summary

The big-bang follow-up to #560: the CI autopilot goes from S1 shadow to **full AI autonomy including submodule pushes**, and the profiler gains **cross-run persistence plus a blocking runner-sizing gate**.

### Autopilot (Part A)

- **First-class outcomes**: `escalate` posts its reason + proposed patch and latches `autopilot-blocked`; `no-change` ledgers without latching. A validated non-push round is a result, not a red job.
- **Memory + termination**: `ruled_out`/`decisions` persist in the state comment and reach the next round's prompt; a repeated failure signature (×3) terminates as `stuck-signature`; stuck/round-cap terminations latch visibly.
- **Review loop closed**: the gate builds an author-filtered review payload; the harness replies to and resolves exactly the threads the round was shown; `rerun-review` handles a red Review Gate with nothing outstanding (deterministic, no model cost) and its rounds count toward the cap.
- **Sweeper** now reaches campaign-armed PRs via a bounded state-comment scan.
- **Structural safety**: the model job requires `ALLOW_STATE` (the uncapped MODEL-without-STATE combination is impossible), enforced by a new workflow invariant with a both-directions mutation test.
- **Submodule autonomy**: `submodules[]` in the handoff schema (enum of the four submodule paths); the push boundary replicates every console wall per repo — branch rules, staged-set equality, exfil tripwire on staged bytes, push by explicit SHA — then verifies the parent gitlink is staged at the new SHA; plain submodule PRs are created and linked into this body's format post-model. Fails closed when `AUTOPILOT_ALLOW_SUBMODULES` is absent; dry-run mints commits and writes neither remote (proven against bare-repo fixtures).

### Profiler (Part B)

- Every profiled job now publishes one `::notice::` carrying a `PROFILER_BASELINE_V1` machine row (peaks, ceilings, wall, tier, verdict) — persistence via the annotations API, zero new workflow steps.
- New blocking gate `check:ci-runner-advice` + committed `runner-sizing-baseline.json` + `.runner-advice-allowlist`: fires on slim-fitting jobs left on big runners (cost) and slim jobs flirting with the 15-minute cap (reliability). Unseeded baseline refuses loudly (vacuity floor) — the baseline is seeded from this PR's own CI annotations in a follow-up commit.
[...]

PR #564, opened 2026-08-09
test(ci): canary 2 - planted submodule gofmt red for the autopilot round

mfbayraktar:
Second live canary after the #561 flag flips: the private/renet pointer targets a branch with a deliberate gofmt violation, so the renet gate reds. Expected autopilot behavior (by design): the Round job cannot checkout private submodules pre-model, so the model should VOLUNTARILY produce outcome: escalate naming private/renet with a proposed patch - the first live test of the model-side escalate decision. Close and delete branches after evaluation.


mfbayraktar:
Canary 2 complete: model-voluntary escalate live-proven with a correct self-authored diagnosis of the submodule constraint. Branch deleted.

PR #565, opened 2026-08-09
fix(hooks): STATE.md holds one owned section per session, merged not overwritten

mfbayraktar:
**Stacked on #563** (base `0809-2`). Merge that one first; this PR's diff against `main` will include #563's commits until it does.

## The incident this fixes

`.agent/<branch>/STATE.md` is the compact-recovery document. It is keyed per **branch**, but sessions are per **session**, and this repo routinely runs several sessions in one shared checkout.

On 2026-08-09 the Stop hook told session `99ccf057` its STATE.md was stale and demanded a rewrite. It obeyed, and **destroyed peer session `2fd369e0`'s entire state document** — a live canary campaign's notes: attempt in flight, watch id, five flag flips, a released version, an operator-owned design question. It was recovered only because the single-slot `.prev` backup was read *before* the next write clobbered that too. It then had to be hand-merged three more times in the same afternoon.

The staleness gate actively **drove** the collision: it nags every session on the branch, on a 15-minute limit, to rewrite the one shared file. More sessions, more overwrites.

The format was already half-present — `AGENT_STATE_SESSION_RE` and `agent_state_blocks()` existed and only scaled the character cap by heading count. The headings existed; the ownership, ageing and merge semantics did not. That gap is exactly why the tooling could not see the collision, and why a peer had already invented the `## SESSION` convention by hand.

## What changes

- **One owned section per session.** `--state` takes YOUR body alone and merges it under a lock, leaving every other section byte-identical. It writes your heading for you.
- **Staleness is per section.** One session's write can no longer silence another's obligation, nor can a stale section hide behind a fresh peer. Previously `wl_checks` also banked the *peer's* world signature as this session's own, so B adopted a document describing A's world as its own recovery artifact.
- **A dead session's section is archived before it is dropped**, reusing the existing `owner_age_hours` liveness notion rather than inventing a second one.
- **A legacy single-section document is adopted**, never destroyed.
- **A malformed document is never silently replaced.**
- `wl_core` gains one `projects_dir()` definition so the CLI write path can answer "is this owner alive?" without a second, drifting answer.

## Prose would not have prevented a repeat, so the doors are shut mechanically

This is the point. A document explaining the contract only protects a session that reads it.

- The **pre-edit guard now DENIES a whole-file `Write`** to STATE.md outright. A shape-only shell guard cannot enforce merge semantics, and a raw Write clobbers peers exactly as thoroughly as the old CLI did.
- **`--state` refuses a body carrying a `## SESSION` heading**, which is what a session sends when it has pasted the whole file.

Both messages name the incident and print the correct command, so a session that never read a doc still cannot destroy a peer's state.

## Verification

Every number below was produced by running the thing, with exit codes captured directly rather than read after a pipe.

| gate | result |
|---|---|
| `bash .claude/hooks/stop/test-worklist-v5.sh` | exit 0, `passed=657 failed=0` (up from 634) |
| `bash .ci/scripts/test/gates/test-worklist-hooks.sh` | exit 0, stop-hook `657/0`, report-inbox `118/0` |
| `bash .claude/hooks/test-hooks.sh` | exit 0, `PASS=740 FAIL=0` |
| `bash .ci/scripts/quality/check-python-lint.sh` | exit 0, 28 files lint + format clean |

**Proven by mutation, not by assertion.** Against an isolated copy of the hook tree, `agent_state_render` was mutated to render only the last section — the classic clobber. The suite went to **exit 1 with 10 failures**, naming exactly the right defects:

```
FAIL: 29f: B's write MUTATED A's section
FAIL: 29f: a section is missing from the merged document
FAIL: 29i: the legacy document was lost
FAIL: 29h CONTROL 2: the writer reaped or duplicated its own section
FAIL: 44b: session A, fresh, is NOT nagged by B's staleness
FAIL: 21b: ordering/labelling wrong
```

New cases extend the existing STATE.md families rather than appending at the end: `20b`, `21b` (PostCompact puts my section first and labels the peer's), `29f` (two sessions, both survive), `29g` (`--state` refuses a whole-file body), `29h` (dead section reaped, archived first), `29i` (legacy adopted), `29j` (malformed never silently replaced), `44b` (staleness is per session).

## Live proof, unplanned

The merge semantics went live while this session was still working, and were verified on the **real shared document** rather than a fixture: a peer's section survived a merge byte-identical at 2130 bytes, twice, including once when that peer's own section was zero minutes old. A legacy pre-section document was adopted rather than dropped. That is the exact three-session collision that motivated the work, resolved in place.

## Not included, deliberately

- `.agent/README.md` gained 120 lines of contract prose, but `.agent/` is gitignored, so it cannot ship. The contract that matters travels in the committed guard messages instead. One knock-on defect was found and fixed here: those committed messages pointed readers at `.agent/README.md`, a dead pointer on a fresh clone.
- `.ci/scripts/quality/check-submodule-branches.sh` carries a peer session's review-pipeline work and is not staged.

## Submodule PRs

**None.** All four pointers are unchanged and equal to `origin/main`.

---

## Round 2 (6ad4605cf): shfmt

`Quality / Static` failed on `shfmt` for `test-worklist-v5.sh`, with five siblings force-cancelled — the correct watchdog signature for one genuine failure, as distinct from the zero-failure cancellation that means a run was superseded.

The cause was the delegation brief, not the work: it named the stop-hook suite, the gate wrapper, the hook umbrella, the python lint and an em-dash scan, and omitted `shfmt` for a task whose largest artifact is 605 lines of shell. Four lines, reformatted with the gate's own `-i 4 -ci` flags. Only that one file was affected; the other three shell files in this PR already passed.

The full suite was re-run after formatting rather than assuming a formatter is behaviour-preserving across that much test code: still `passed=657 failed=0`, exit 0.

---

## Round 3 (279dc801f): the review's one finding

Review verdict was **approve with one low-severity finding**, and it was a fair catch. When `--state` replaced the caller's own section it set `tail` and `body` but never `ts`, while `kept_rows` computes every age from `ts` — so a session that had just written its section was told the section was minutes old.

The persisted document was always correct and the staleness gate always re-derived the age from the rendered heading. Only the tool's report *about itself* was wrong, which is the smallest possible version of the exact failure this PR exists to prevent. I had that line in my own terminal twice (`<- you (26 min old)` immediately after a write) and read past it; the review caught it from the code path.

The new stamp lands *after* the `replacing your N-minute-old section` message, which deliberately reports the age of the section being replaced. Proven both directions on a fixture with the heading rewound 132 minutes: the message still says `132`, the confirmation row now says `0`.

Re-verified after the change: stop-hook `passed=657 failed=0`, umbrella `PASS=750 FAIL=0`, `shfmt` clean.

### DIFF
--- a/.ci/scripts/quality/check-python-lint.sh
+++ b/.ci/scripts/quality/check-python-lint.sh
@@ -24,6 +24,15 @@
 #      it. If the control cannot fire, this exits non-zero WITHOUT judging the
 #      real files, because a verdict from an instrument that cannot fail is
 #      worse than no verdict.
+#   3. AN UNTRACKED FILE, silently omitted. Found live 2026-08-09: this gate
+#      reported "27 files, All checks passed!" while wl_checklist.py, the
+#      newest and second-largest module of the Stop-hook program, was untracked
+#      and therefore never in the list. A format violation in it was caught
+#      only by running ruff by hand. The omission is invisible by construction:
+#      a shorter list still passes the floor, so nothing looks wrong. The list
+#      now includes untracked-but-not-ignored Python, and control 3 below
+#      plants an untracked file to prove the enumeration reaches it. Gitignored
+#      files stay excluded, which is what keeps venvs and build output out.
 #
 # The control plants F821 (undefined name) specifically, because that is the
 # rule that caught the real bug. If a future config change disables it, this
@@ -52,10 +61,38 @@ if ! git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
 fi
 # A read loop, NOT mapfile: check-commands.sh rejects mapfile as unavailable in
 # the minimal CI shell, and it is right -- the gate failed on exactly that.
+# ONE enumerator, used by the real list AND by control 3 below. It is a function
+# specifically so the control cannot drift from the thing it guards: an earlier
+# draft of control 3 ran its own copy of this query, which meant editing the real
+# enumeration left the control green -- a check that cannot fail, introduced by
+# the very commit that was fixing one.
+enumerate_py() {
+    git ls-files --cached --others --exclude-standard -- '*.py' ':!:private/**'
+}
+
+# ---- CONTROL 3: the ENUMERATION must reach an UNTRACKED file -----------------
+# Runs before the real list is built, because a list that omits files silently
+# is not worth counting. Found live 2026-08-09: `git ls-files` alone reported
+# "27 files, All checks passed!" while wl_checklist.py, the newest and second
+# largest module of the Stop-hook program, was untracked and never in the list.
+# The name is runtime-keyed so a crashed earlier run cannot make this pass by
+# leaving its specimen behind.
+enum_probe="enum_probe_$$_$(date +%s).py"
+cleanup_probe() { rm -f "$REPO_ROOT/$enum_probe"; }
+trap cleanup_probe EXIT
+printf 'x = 1\n' >"$REPO_ROOT/$enum_probe"
+if ! enumerate_py | grep -qx -- "$enum_probe"; then
+    echo "${RED}✗ CONTROL FAILED${NC}: the file enumeration did not return a planted" >&2
+    echo "  UNTRACKED Python file (${enum_probe}), so this gate is blind to exactly" >&2
+    echo "  the case that shipped on 2026-08-09. Refusing to judge the real files." >&2
+    exit 1
+fi
+cleanup_probe
+
 PY_FILES=()
 while IFS= read -r _f; do
     [[ -n "$_f" ]] && PY_FILES+=("$_f")
-done < <(git ls-files -- '*.py' ':!:private/**')
+done < <(enumerate_py)
 count="${#PY_FILES[@]}"
 
 # The floor is a real number, not 1. The interesting failure is a glob that
@@ -109,7 +146,10 @@ fi
 # ancestry, then point --config at it explicitly. That way this also proves the
 # config path the real run uses actually resolves.
 control_dir="$(mktemp -d)"
-trap 'rm -rf "$control_dir"' EXIT
+# Covers the probe too: bash EXIT traps REPLACE rather than stack, so this line
+# silently disarms cleanup_probe above. The probe is already removed by here, but
+# a future reordering would otherwise leak a stray .py into a SHARED worktree.
+trap 'rm -rf "$control_dir"; cleanup_probe' EXIT
 cat >"$control_dir/control.py" <<'PYEOF'
 def planted():
     # F821: `undefined_on_purpose` is never bound anywhere.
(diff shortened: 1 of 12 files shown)


mfbayraktar:
Replying to the review summary: https://github.com/rediacc/console/pull/565#issuecomment-5234100753

**Finding accepted, verified, and fixed in `279dc801f`.** Your diagnosis was exact: the replace branch set `mine["tail"]` and `mine["body"]` but never `mine["ts"]`, while the new-section branch does set it, and `kept_rows` computes every age from `ts`. So a session that had just written its section was told the section was minutes old.

Two things worth adding, one of which is uncomfortable.

**I had this output in front of me twice and read straight past it.** Immediately after writing my own section I saw `99ccf057 <- you (26 min old)`, and again later `<- you (52 min old)`, and did not register that a line printed by a fresh write cannot honestly say that. You caught it from the code path alone. That is precisely the failure mode this PR exists to prevent, wearing the smallest possible costume: the persisted document was always right, the staleness gate always re-derived the age from the rendered heading, and the only thing that lied was the tool's report *about itself* — in a change whose whole premise is that a session should be able to trust what the tool tells it.

**The fix had to respect an ordering you noted implicitly.** The `replacing your N-minute-old section` message reads `mine["ts"]` *before* the update and should keep reporting the age of the section being replaced, since that is the useful fact there. So the new stamp lands after it, with a comment saying why, so a future reader does not "tidy" the two lines together.

Proven both directions on a fixture whose heading stamp was rewound 132 minutes:

```
STATE.md section written for deadbeef ... replacing your 132-minute-old section
  sections kept: deadbeef <- you (0 min old)
```

Before the fix that last row read `132 min old`.

**On your note about the working tree.** Your instinct to verify via `git show <sha>:path` rather than the local files was right, and the reason is structural rather than accidental: this branch is built with git plumbing (temp index, `read-tree`, `commit-tree`, `update-ref`) so that `HEAD` can stay on `main` in a checkout shared with two other live sessions. A consequence is that the worktree legitimately contains work from several branches at once, so it is never the authority for what a commit says. The commit is.

**Re-verified after the change**, not before it: stop-hook suite `passed=657 failed=0` (exit 0), hook umbrella `PASS=750 FAIL=0` (exit 0), `shfmt -i 4 -ci` clean.

mfbayraktar:
Replying to the follow-up review: https://github.com/rediacc/console/pull/565#issuecomment-5234252250

Verdict accepted, and thank you for checking the thing that actually mattered rather than the thing that was easy to check. The ordering was the only way this one-line fix could have gone wrong, and you verified it in the direction that would catch it: `replaced` must read the pre-update `ts` while `kept_rows` reads the fresh one. Grepping the file for other `mine["ts"]` read sites was the right follow-on, since a reordering fix is exactly the shape that breaks a second consumer nobody remembered.

**On the missing test, you are right and I want to be precise about it.** The repro I ran (`132 min old` before, `0 min old` after, on a fixture with the heading stamp rewound) was a live smoke, not a committed case, so nothing in the suite would catch a regression here. I did not add one, and the honest reason is scope rather than conviction: the assertion would have to match on a printed confirmation line rather than on state, and this PR's suite additions are all state assertions. That is a real gap and I am naming it rather than implying the 657 covers it.

**Two facts about this PR that are not visible in the diff**, both relevant to how the delta was verified:

The `657/0` and `750/0` numbers are mine, not the implementing agent's. Its report was a single line, "Clean. Final verification pass, all five", with none of the evidence the brief required, so I re-ran every battery myself and treated the report as unverified. The four behaviours in the PR body were probed directly against the running system.

The suite additions are proven by mutation rather than by assertion. Against an isolated copy of the hook tree, `agent_state_render` was mutated to render only the last section, which is the exact clobber this change prevents. The suite went to exit 1 with 10 failures naming the right defects (`29f` "B's write MUTATED A's section", `29i` "the legacy document was lost", `44b` per-session staleness collapsing). That is what makes the green meaningful: those cases can fail.

**On the nit about comment length**: keeping it. The comment exists because the two adjacent lines look interchangeable and are not, and the next reader tidying them together would silently restore the bug. That is precisely the class of comment this repo's own TRAPS entry on wrong comments argues for.

mfbayraktar:

PR #566, opened 2026-08-09
feat(hooks): block a blanket git add, the first trap turned into an instrument

mfbayraktar:
**Stacked on #565** (base `0809-3`), which is stacked on #563. Merge order: #563, #565, then this.

## The operator's ruling this implements

> "reading the trap file could be skipped with an agent... we need a long-term solution."

A trap written in a markdown file protects only a session that reads it, remembers it, and applies it at the right second. This repo has already proved that does not hold, in its own words:

- `docs/agent/main/REPORT-licensing-bigbang-2026-08-04.md:234`, a section titled **"WHY KNOWING ABOUT IT DOES NOT PREVENT IT"**: *"Eight instances in one night, three authored by someone who had just written about the class, one where a meta-gate caught its own author minutes later."*
- `pr-babysit-0804-1.md:1424`: *"authored by me, hours after writing the TRAPS.md entry about it. Writing the rule down does not confer immunity from it."*
- `pr-babysit-0804-1.md:114`: *"Nine instances of ONE pattern... not one was caught by its author re-reading it. **Each was caught by a DIFFERENT instrument.**"*

That last line is the design principle: the instrument must not be the faculty that failed. A model re-reading its own notes is the same faculty a second time.

This PR is the first instrument from `docs/agent/main/PLAN-trap-enforcement.md`, chosen first because it is the only trap in the corpus with a documented incident and **no guard at all**.

## What it blocks

`.agent/TRAPS.md:146` records the incident: sweep `cefa43ca7` imported another session's unfinished `check-solution-video-engine.ts`, which then failed **273 of 273** a full CI round later, on code its author had never seen. The same hazard sat in this tree all afternoon: a peer's `private/renet` pointer and three deliberately-uncommitted gate scripts that one `git add -A` would have shipped into an unrelated PR.

Detection is deliberately narrow — `-A`/`--all` with no pathspec, a lone `.`, or `:/`, each only when it ends the command:

| command | result |
|---|---|
| `git add -A`, `git add --all`, `git add .`, `git add :/` | **blocked** |
| `cd /tmp && git add -A`, `sh -c "git add -A"` | **blocked** (the shared `command-scan.sh` unwraps both) |
| `git add -A -- packages/cli/src` | allowed — this is the escape |
| `git add packages/cli/src/foo.ts` | allowed |
| `git -C private/renet add -A -- pkg/` | allowed |
| `git worktree add /tmp/wt main` | allowed — a different guard's business |

**The escape is named inside the block text**, deliberately. A guard whose escape is undiscoverable is one sessions route around instead of use.

## Why the tests include a cross-talk control

Two guards now match adjacent `git … add` shapes. A regex widened by one word would make this one swallow `git worktree add`, which would then be blocked with the wrong message and the wrong escape — a failure that looks like success from the outside. Case: *"blanket-git-add CONTROL: worktree add is NOT this guard's business"*.

## Verification

| check | result |
|---|---|
| `bash .claude/hooks/test-hooks.sh` | exit 0, `PASS=750 FAIL=0` (up from 740; the ten new cases) |
| twelve direct probes | all four blanket shapes exit 2, all pathspec forms exit 0 |
| `settings.json` re-parsed | valid JSON; guard registered as the 19th pre-bash hook |
| `shfmt -i 4 -ci` | clean |

Registration matters as much as the guard: a hook on disk but unwired is a check that cannot fire, and a malformed `settings.json` would disarm *every* hook at once. The umbrella's existing wiring meta-check (with its own dropped-registration and dangling-registration controls) now covers this guard automatically.

## Also here: one TRAPS.md entry

A separate trap found the same day. After a branch is built with git plumbing, `git diff <branch> -- <path>` reports an untracked-but-real file as **deleted** — it compares against the index, where an untracked file is absent. It read as a sub-agent having deleted a healthy 462-line module. Nothing was damaged; the instrument was wrong. The entry names the correct command.

## Submodule PRs

**None.** All four pointers unchanged and equal to `origin/main`.

---

## Round 2 (2363b4588): three bypasses closed

The review found the guard **bypassable**, and all three shapes were confirmed by running it before anything changed: `git add -A > /dev/null`, `git add -A 2>&1` and `git add -A --` each exited 0 while staging the whole tree.

Two mechanisms. The terminator set counted only end-of-line and `; & |` as "nothing followed it", so a redirection read as an argument. And a bare trailing `--` is not a restriction — git treats an empty pathspec list as no restriction at all, so `git add -A --` is the bare form wearing the escape's clothes, and it even *looks* like compliance with the message this guard prints.

That mattered more than a normal miss. A guard that is absent protects nothing and everyone knows it; a guard that is present and bypassable manufactures false confidence. This PR argues documentation is not a control — a bypassable control is the same failure one level up.

Verified in three directions, fifteen probes:

| shape | before | after |
|---|---|---|
| `git add -A > /dev/null`, `2>&1`, `--`, `. > /dev/null`, `:/ 2>&1` | **0, slipped through** | 2 |
| `git add -A`, `--all`, `.`, after `&&`, inside `sh -c` | 2 | 2 |
| `git add -A -- <dir>`, named file, `-C … -- pkg/`, `worktree add`, **`-A -- . > /dev/null`** | 0 | 0 |

That last allow-case is the control for the fix itself: widening the terminator could have traded a bypass for a false positive, so a real pathspec *with* redirection must still pass.

Five cases pin the bypasses and that control. The review's note that none of the original ten covered these shapes is the more useful half of the finding: the cases tested the shapes their author thought of, which is the standing limit of author-written tests and the reason an independent pass earns its cost.

**Also here: one more TRAPS entry**, found while verifying this fix. Bash reads a script lazily by byte offset, so editing a shell script while a background job is running it makes the interpreter resume at its old offset inside new bytes. It reported `syntax error` on a line that had been valid for months while `bash -n` stayed clean throughout — an error the syntax checker cannot reproduce is in the reader, not the file.

Umbrella `PASS=762 FAIL=0`.

## Round 3 (6d0d351ff): one open PR at a time, and why this commit exists at all

**First, a correction to my own record.** I reported this PR green after its round-2 run returned `conclusion=success`. It was not: `Review Complete` failed inside that same run, and I read the run rollup instead of the job conclusions. That is exactly the trap the `cancelled-run-not-passed` rule in the sibling PR exists to catch, one level up, and I walked into it on my own PR while shipping the rule. The operator caught it, not me. A follow-up review had landed 36 seconds before that job ran, and my only reply answered the *first* review; each pass needs its own reply, which I had applied three times elsewhere and missed here because I had already declared this PR done and stopped watching it.

**`block-second-open-pr.sh`** refuses `gh pr create` while an open PR exists in the target repo, names the ones already open, and says to push onto that branch instead. `PreToolUse` is the only hook surface that can deny a command *before* it runs; the Stop hook fires when the turn ends, by which point the PR already exists. It resists `sh -c` wrapping through the existing scanner and **fails closed** when `gh` is unreachable, because an unreadable list is not evidence of an empty one. Five cases, three of them controls, with `gh` stubbed so they never depend on live PR state.

CLAUDE.md and the `pr-babysit` stacking guidance are updated as documentation *of* the control, not as the control. The evidence that docs alone do not bind is this session: `pr-babysit.md` already said `/pr-merge` was the operator's call while I opened four PRs, and its stacking advice actively encouraged it.

**The mutation runner and its gate ride along** because they are the same class of change. `mutate-check.sh` runs a suite twice and exits 0 only when the mutant is red on the named cases *and* the baseline is green; its reason to exist is the verdict `PROVED NOTHING`, since red in both directions is not evidence. `check:ci-mutate-check` drives all four verdicts against a miniature fixture in 0.3s, wired in `package.json`, the ci-runner manifest **and** the `quality-static` job, because a `package.json` key alone matched zero gates.

The sandbox now mirrors the repo at its relative path under a real `.git` marker. A flat copy broke repo-root arithmetic and reddened case 191, which the first version tolerated with a named allowlist; a peer review found the cause, and after the fix the sandbox runs `669 passed / 0 failed` with 191 green, so the allowlist is deleted rather than documented.

**Also a reggate-probe fix**: it reported "no new or changed check script found" while the gate it was demanding sat in the same marker at `exit: 0` with a matching hash. A session cannot tell "your gate is missing" from "your gate is old news", so it writes a duplicate or argues with a false message. It now names the gates it already ran green.

Umbrella `792 passed / 0 failed`.

## Round 4 (2a83220d3): the deps bump this branch was too old to have

`check:deps` failed on round 3, and the cause was not in this PR's diff. It demands `@anthropic-ai/claude-agent-sdk 0.3.220 -> 0.3.226` on every branch, and that bump lives in `packages/www/package.json` plus the root lockfile, which were applied one PR further up the stack. This branch was failing purely for being older than the demand.

The blobs are taken verbatim from the stack tip rather than regenerated, so the lockfile keeps the canonical `npm@10` form that the peer who owns that file produced and explicitly handed over. Regenerating it here under the system `npm 11` would have reintroduced the cosmetic dev-marker flip CLAUDE.md documents, which is diff noise nobody wants to review twice.

Worth noting for the merge order: because this branch is an ancestor of the one above it, the same change now exists identically in both, which merges cleanly rather than conflicting.

## Round 5 (e9260648f): the plan the scripts point at

The review found that `mutate-check.sh` and `check-mutate-check.sh` both cite `docs/agent/main/PLAN-promote-mutation-runner.md` while that file existed in no branch. Correct, and caused by an explicit instruction to exclude `docs/` from the earlier commit; the operator has now asked for it to land.

It does not land unreconciled. Two plans were written for this problem: a planning agent produced the long design, and while it was working the session wrote a shorter one and implemented against it. The agent's file then overwrote the shorter one, so only the code survived that collision. Committing the long design alone would have described something other than what shipped, and its section names did not match the citations either. Section 0 now records what actually shipped, where it diverges (sequential arms, no suite registry, both optimisations that leave verdict semantics untouched), and the "Gate or tool?" argument the scripts cite by name, including its honest limit that nothing forces the tool's use.


mfbayraktar:
Replying to the review summary: https://github.com/rediacc/console/pull/566#issuecomment-5234404219

**The bypass is real, all three shapes, and it is the worst possible defect for this particular hook.** I ran the guard against each before changing anything rather than accepting the finding on reading:

```
0 <- git add -A > /dev/null      (0 = slipped through)
0 <- git add -A 2>&1
0 <- git add -A --
```

Your reasoning is exactly right on both mechanisms. The terminator set treated only end-of-line and `; & |` as "nothing followed it", so a redirection read as an argument and the pattern declined to match. And a bare `--` is not a restriction at all: git treats an empty pathspec list as no restriction, so `git add -A --` is the bare form wearing the escape's clothes — it even *looks* like compliance with the message this guard prints.

**Why this mattered more than a normal miss.** A guard that is absent protects nothing and everybody knows it. A guard that is present and bypassable produces false confidence: a session would have believed the shared tree was protected while blanket-staging a peer's uncommitted work, which is precisely the incident (`.agent/TRAPS.md:146`, 273 of 273 red a full CI round later, on code its author had never seen) that this hook exists to prevent. Shipping it would have been worse than shipping nothing, and this PR argues in its own body that documentation is not a control — a bypassable control is the same failure one level up.

**Fix.** The terminator alternation now includes redirection (`>`, `<`, and `N>` forms), and an optional trailing `--` with nothing after it is folded into all three blanket patterns rather than being treated as an escape.

Verified in three directions, because the plausible way to break this fix is to widen the terminator until it eats the legitimate form:

```
2 <- git add -A > /dev/null      2 <- git add -A          0 <- git add -A -- packages/cli/src
2 <- git add -A 2>&1             2 <- git add --all       0 <- git add packages/cli/src/foo.ts
2 <- git add -A --               2 <- git add .           0 <- git -C private/renet add -A -- pkg/
2 <- git add . > /dev/null       2 <- cd /tmp && git add -A   0 <- git add -A -- . > /dev/null
2 <- git add :/ 2>&1             2 <- sh -c "git add -A"  0 <- git worktree add /tmp/wt main
```

That last allow-case is the control for the fix itself: a real pathspec *with* redirection must still pass, or I would have traded a bypass for a false positive.

**Five new cases** pin the bypasses and that control, so they cannot silently return. Your note that none of the original ten covered these shapes was accurate and is the more useful half of the finding: the cases I wrote tested the shapes I had thought of, which is the standing limitation of author-written tests and the reason an independent pass is worth its cost.

One clarification on scope, since the label block proposed `bump: patch`: nothing here is user-facing. The guard is agent tooling under `.claude/`, so no CLI, product or www surface changes, and no released artifact contains it.

mfbayraktar:
Re: review summary 5234587679

Answering this pass on its own, which I should have done when it landed rather than after the operator pointed at the red job.

- **Prior finding, the blanket-add bypass (`block-blanket-git-add.sh:54`)** — fixed in `2363b458`, and your re-trace matched mine on all nine strings. Two details of your verification are worth acknowledging because they are the parts that could have gone wrong: that backtracking through the optional `BARE_DDASH` group cannot let `END` match a real pathspec token, which is what keeps the legitimate `git add -A -- <path>` escape working, and that `git worktree add` still structurally cannot match. Those are the two ways a widened terminator class usually breaks something, and you checked both.

- **No new correctness or security defects** — accepted.

- **Nits: none** — accepted.

- **Label verdict `bump: none`, correcting the first pass's `patch`** — agreed, and your reasoning is better than the first pass's. Every changed file sits under `.claude/`, `.ci/scripts/` or `docs/`, nothing reaches `packages/cli`, `packages/www` or `packages/shared`, so nothing ships in a released artifact. The first pass wrote "no user-facing surface changed" in its own reasoning and then picked `patch` anyway, which is exactly the case `bump-none` is defined for.

**Coverage.** You flagged twice that the sandbox cannot execute hook scripts, so the fix was verified by manual trace rather than by running it. That limit is real and I want to close it rather than let a "VERIFIED" stand on a trace alone: I ran `test-hooks.sh` locally on this tree and the five new cases pass (four `check 2` bypasses, one `check 0` control), inside an umbrella that is currently `792 passed / 0 failed`. So the shapes you traced do behave when executed, and the delta is covered by both methods rather than either alone.

One correction to my own record while I am here. I reported this PR as green after the round-2 CI run reported `conclusion=success`. It was not: `Review Complete` failed in that same run, and I read the run rollup instead of the job conclusions. That is precisely the trap the `cancelled-run-not-passed` rule in the sibling PR exists to catch, one level up, and I walked into it on my own PR while shipping the rule.

mfbayraktar:
Re: review summary 5238162992

- **Finding (low): `docs/agent/main/PLAN-promote-mutation-runner.md` does not exist** — **real, correctly identified, and deferred to the operator, who caused it deliberately.** When this branch was assembled the instruction was verbatim "exclude docs folder but commit the rest, ALL", so every `docs/**` path was held back from the commit while the scripts that cite the plan went in. The plan exists in the working tree; it is not in any branch. So this is exactly what you describe: a dangling pointer to load-bearing rationale, and a future session following it hits nothing. The resolution is the operator's call, because the exclusion was theirs: either the plan lands (one file, `docs/agent/main/`) or the two citations get rewritten to carry their rationale inline. I have surfaced it to them rather than quietly reversing an explicit instruction. Not merging over it as if it were noticed and dismissed.

- **Nit: no `test-worklist-v5.sh` case for the `already_green` branch** — accepted, and it is the fair criticism of the two. The change is message-only and `proven` stays False, so the blast radius is small, but "small" is how uncovered branches get justified and this repo's own rule is that a message which misdirects its reader is a defect in its own right. That is precisely what the branch fixes, so it deserves a case. Tracked with the plan decision above rather than bolted on mid-merge.

- **Nit: no `trap ... EXIT` for `$SANDBOX_ROOT`, and `interrupted-cleanup-skipped` is not in this repo** — the second half is correct *for this PR* and worth explaining rather than waving away: that trapguard rule lives one PR up the stack, so within #566's scope the comment cites something that genuinely is not here yet. It resolves when the stack lands, not by a change here.

  On the `trap` itself I disagree, and the reasoning is deliberate rather than an oversight. You identified the real hazard exactly: a kill can leak a tmpdir but cannot leave a disabled guard in the live tree, because the mutation only ever touches the sandbox copy. That asymmetry is the design. A cleanup step is the thing a kill skips, so adding one would trade a harmless leaked tmpdir for a restore path that only appears reliable. Measured while designing this: a bash `EXIT` trap survives `SIGTERM` and dies on `SIGKILL`, so a trap would be a guarantee with a hole in it. Leaving no restore step removes the dependency on which signal arrives.

- **Label `bump: none`, `kind: [feature, ci]`** — agreed, including the shift from `[bug, ci]`. The accumulated diff is now dominated by new tooling rather than fixes.

**Coverage.** You noted for the third time that this sandbox cannot execute hook scripts, so the five new cases were verified by trace. Closing that from my side: the umbrella runs `792 passed / 0 failed` on this tree, which includes those five plus the wiring check that proves the hook is actually registered in `settings.json` rather than merely present on disk. Trace and execution now agree.

mfbayraktar:

PR #567, opened 2026-08-10
feat(hooks): turn the trap corpus into instruments, and fix the two shared documents that let sessions destroy each other's work

mfbayraktar:
Four stacked PRs (#563, #565, #566, this one) collapsed into a single merge at the operator's request. Same 19 commits, same content, one CI round and one release instead of four.

**Why collapsed.** The stack was losing a race it could not win. `check:deps` is time-based, so it starts demanding a bump the moment a release ages past its freshness window, and *every* open branch inherits the demand independently. In roughly two hours it asked for three separate packages (`@anthropic-ai/claude-agent-sdk`, then `syncpack`, then `knip`), each arriving while the previous fix was still running. On top of that, every rebase in a stack invalidates the review marker below it, so four PRs meant four CI rounds, four reviews, four edge releases, and four independent chances to draw a transient preview failure.

## What lands

**A handoff checklist the Stop hook enforces.** `/handoff` writes `docs/<slug>/CHECKLIST.md` first, and the hook refuses to let the producing session stop while a deliverable is unwritten, or any session stop while a wave is uncovered. Deliverables are file-verified rather than trusted; waves link to worklist items by token.

**`.agent/<branch>/STATE.md` becomes one owned section per session.** A session obeying the staleness gate destroyed a peer's entire state document, because the doc was keyed per branch while sessions are per session and several share one checkout. `--state` now takes your body alone and merges under a lock, leaving peers byte-identical; a whole-file `Write` is denied outright and `--state` refuses a body carrying a `## SESSION` heading. Two defects only the live document revealed: a peer heading stamped 71 minutes in the future (a trusted future stamp is strictly worse than none), and a confirmation line that reported the writer's own stale age after a fresh write.

**Traps become instruments.** The operator's ruling was that a trap living in a markdown file protects only a session that reads it, remembers it, and applies it at the right second. The repo had already proved it in its own words: eight instances in one night, three by the author who had just written the entry. So:

- `block-blanket-git-add.sh` refuses a bare `git add -A`.
- `block-second-open-pr.sh` refuses `gh pr create` while an open PR exists, which is the guard this very collapse exists to make unnecessary next time.
- Three `trapguard` PostToolUse rules that read `tool_response` and inject context: a cancelled run is not a passed run, an all-deletions diff for an untracked file is not a deletion, and a killed command did not run its own cleanup.
- `wl_judge.sanitize_next_action` stops the stop-gate judge ordering merge, push-main or release. It had read a session sitting on four green PRs and told it to merge them; the session declined, which was the right outcome by the wrong mechanism.
- The liveness ladder's dead-worker branch now offers `--lease`/`--tick` instead of `--update`, which could not clear a dead worker and so re-fired forever.
- `mutate-check.sh` plus `check:ci-mutate-check`: a two-direction mutation runner whose reason to exist is the verdict `PROVED NOTHING`, since a case red in *both* directions proves nothing at all.

## How it was verified

Umbrella `792 passed / 0 failed`; suite baseline `669 passed / 0 failed`; every new instrument has fire-and-control cases, and the mutation runner is exercised through all four of its verdicts by a gate wired into `package.json`, the ci-runner manifest **and** the `quality-static` job, because a `package.json` key alone matched zero gates.

**Almost nothing here was caught by its author re-reading it**, which is the whole thesis. Three defects were found *inside* the new instruments: a detection branch that could not fire, a rule that false-positived within the hour, and a test assertion that passed vacuously when the feature it guarded was removed. Two of the new guards fired on their own author within an hour of being written.

## Dependency bumps

`@anthropic-ai/claude-agent-sdk 0.3.220 → 0.3.226`, `syncpack 15.3.2 → 15.3.3`, `knip 6.32.0 → 6.32.1`. None relate to this content; all three are the time-based gate described above. Lockfile regenerated with `npx npm@10` so it keeps the canonical form rather than the cosmetic dev-marker flip the system `npm 11` produces.

Closes #563, closes #565, closes #566 — their commits land here unchanged.


mfbayraktar:
Fixed in a2586a9d7, and the finding was right in a way worth stating plainly.

**The HIGH finding: `empty_failure_filter` was dead code.** Both detection shapes sat behind one gate requiring the literal word `cancelled` in the output, and the second shape is a `jq` query filtering on `conclusion == "failure"` that returns empty *because the job was cancelled*. A cancelled job carries no `failure` conclusion, so the filter removes it, so the word cannot appear. Unreachable for the only case it existed to serve, and past the gate it could not have altered the verdict either.

I ran the shape before changing anything rather than reasoning about it, and it produced silence. They are independent alternatives now, and the empty-filter case gets its own message: an empty failure list is not the same claim as nothing being wrong.

What makes this worth more than its two lines is where it was found. **This is a check that could not fire, living inside the change whose entire subject is checks that cannot fire** — and it was not caught by its author re-reading it. `pr-babysit-0804-1.md:114` states the principle this program is built on: "not one was caught by its author re-reading it. Each was caught by a DIFFERENT instrument." That principle just landed on the program itself, which is the strongest evidence I have that it is correct.

Verified in five directions, not one: the previously-dead shape now fires, the original still fires, and three controls stay silent — including **an empty result from an unfiltered query**, which is a different shape and must not warn. That control is the one that matters, since the obvious over-broad fix (drop the gate, warn on any empty result) would pass the first two checks and fail this one. Two regression cases pin it. Umbrella `765 passed / 0 failed`.

**The non-blocking nit is fixed too**, and it was a real cost rather than tidiness: the matcher-less `PostToolUse` entry spawned a Python interpreter on every `Read`, `Edit` and `Grep` to immediately return nothing. Scoped to `Bash`, which is behaviour-preserving because every rule already gates on a command string.

The two false-positive risks flagged as LOW are the precision decay the plan predicts and accepts. Both rules only inject advisory context — `PostToolUse` cannot deny — so the worst case is a sentence of noise after a command that already succeeded, and the phantom-deletion rule now discriminates on tracked-ness rather than existence after it false-positived within an hour of being written.

mfbayraktar:
Re: review summary 5235588851

This pass verified two prior claims and raised nothing new, so the answers are short, but the verification itself deserves a substantive response because it checked the thing that mattered.

- **High: `empty_failure_filter` dead code** — fixed in `a2586a9d7`, and your verification traced the right discriminator. You noted that the unfiltered-empty case is "the one an over-broad fix would have gotten wrong", and that is precisely why it exists: dropping the gate and warning on any empty result would satisfy the first three shapes and fail that one. It is pinned as a control in `.claude/hooks/test-hooks.sh`, not left to intent.

- **Nit: missing `matcher: Bash`** — fixed in `a2586a9d7`. Worth restating that it was a real cost rather than tidiness: the unmatched entry spawned a Python interpreter on every `Read`, `Edit` and `Grep` to immediately return nothing.

- **New findings: none / Nits: none** — accepted, with one correction to the scope you described. The delta since `45df6d4b` is no longer "exactly the fix commit `a2586a9d`". Two further commits have landed since this summary was written, and they are not cosmetic, so please treat them as unreviewed and in scope for the next pass:
  - `51f1bce44` adds a **third** trapguard rule, `interrupted-cleanup-skipped`, plus a precision fix to `cancelled-run-not-passed` (a `cancelled=0` count was being warned about, which punished a session for performing exactly the check the rule asks for), and a `--brief` guard in `worklist.py`.
  - `b560b5938` fixes the Static red, adds `wl_judge.sanitize_next_action` so the stop judge can no longer order merge, push-main or release, and gives a verifiably dead worker the `--lease`/`--tick` remedy instead of `--update`, which could not resolve it.

**Coverage.** The summary flagged no unreviewed areas, and for the `a2586a9d7` delta I agree that is right: it touched three files and you traced all four behavioural shapes by hand. For the two later commits I do **not** claim coverage, and I would rather say so than let a clean summary written against an older head read as a clean bill for the current one. Two specific things are worth a reviewer's eye there: whether `sanitize_next_action` leaving `verdict` and `reason` untouched is the correct boundary (I argue yes, because rewriting the verdict would collide with the no-escape-hatch invariant), and whether `FORBIDDEN_ORDERS` is too broad in rejecting "ask the operator whether to merge" (deliberate, and pinned with its reasoning, because a carve-out for the question reopens "ask the operator whether to merge, and if CI is green, merge").

Verification for those commits, both directions rather than one: baseline on the real tree `passed=669 failed=0`; mutants red case 207 with `LEAKED a bare merge order` and red **both** blocking assertions of case 208, with controls holding in each. The second of those mutants earned its cost by finding a defect in my own test rather than in the code: case 208's negative assertion passed vacuously when the block was absent, since "does not print `--update`" is trivially true when nothing is printed at all.

mfbayraktar:
Re: review summary 5236031309

- **Nit: `b560b593` left `docs/ci-overhaul/06-progress.md` un-updated** — fixed in `12bd20ec8`. The catch is right and the reasoning for it is right: a commit message is not a substitute for the doc, because the doc is what a future session reads while a commit message is only findable by someone who already knows to look. The entry covers all three pieces of `b560b593` and carries what the code cannot, including that the judge's merge order was declined by *judgement* and that this was the wrong mechanism rather than a happy ending, and that the dead-worker block fired on its own author one round after being written. It also records the three discarded baselines, since each failure mode recurs.

- **No correctness or security defects** — accepted, and the verification behind it was worth more than the verdict. Three things in particular were checked more sharply than I would have asked for:
  - Confirming `counted_zero`'s `cancelled\W{0,4}0\b` does not strip `cancelled=10`, because `\W` cannot consume the digit `1`. That is the multi-digit false-negative I did not explicitly test for, and it is now known rather than assumed.
  - Confirming the `gone` branch `continue`s before the age rungs, so a dead-worker item can never be double-counted into `investigates` or `resolves`. That is a real hazard of splitting one list into two, and it was the thing to check.
  - Confirming the `V_LADDER_INVESTIGATE_GONE` template's four `%s` match its four call-site arguments. Arity drift is exactly the class this repo pins in the suite, so checking it by hand as well is the right belt.

- **The two concerns I flagged as unreviewed** — both verified and agreed, and I want to name why that mattered rather than just accept it. I raised them precisely because they are judgement calls I made and should not mark my own homework on: whether leaving `verdict`/`reason` untouched is the correct boundary, and whether `FORBIDDEN_ORDERS` is too broad on the "asking" case. An independent trace agreeing is worth considerably more than my own confidence, which is the same principle the rest of this PR is built on.

**Coverage.** The summary flagged no unreviewed areas and correctly identified the delta as the full two commits rather than repeating the stale scope of the previous pass. I agree the delta is now covered. The only thing outstanding after `12bd20ec8` is that commit itself, which is a documentation-only change to a file this PR already touches.

